PR #1, opened 2020-06-25
add maven wrapper to project

isopov:
This will allow to build project without installing maven locally, also it will ease using correct maven version for contributors.

This wrapper is is third-party currently, but it is on process of transitioning to Apache Maven project and will be part of it in the next release.


akudiyar:
@isopov would you mind adding a notion about using mvnw to README?

isopov:
@akudiyar

Added note on runnig tests to README using maven wrapper. Changed links to project (they pointed to previous location on github)

akudiyar (approved):

PR #2, opened 2020-06-25
remove spring dependency

isopov:
Using spring framework only for nullability support and assert utils looks strange. Half of java projects in the wild do not use spring at all and this dependency looks too heavyweight. But another half of projects may have even bigger problems due to this dependency from version conflicts.


isopov:
Кажется, это проще уже заново сделать.

akudiyar:
Created an issue to follow up this PR: #7

PR #3, opened 2020-07-27
Add TarantoolTuple setField by position method

savolgin:
Add TarantoolTuple setField by position method

https://www.notion.so/TarantoolTuple-0530683ae98a41cf9f3f6574c31ba056

### DIFF
--- a/src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
+++ b/src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
@@ -1,6 +1,7 @@
 package io.tarantool.driver.api.tuple;
 
 import io.tarantool.driver.exceptions.TarantoolValueConverterNotFoundException;
+import io.tarantool.driver.mappers.MessagePackObjectMapperException;
 import io.tarantool.driver.protocol.Packable;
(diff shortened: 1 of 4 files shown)


src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
```diff
@@ -27,4 +28,11 @@
      * @throws TarantoolValueConverterNotFoundException if the converter for the target type is not found
      */
     <O> Optional<O> getObject(int fieldPosition, Class<O> objectClass) throws TarantoolValueConverterNotFoundException;
+
+    /**
+     * Set a tuple field by its position
+     * @param fieldPosition the field position from the the tuple start, starting from 0
+     * @param value new field value
+     */
+    <V extends Value> void setField(int fieldPosition, V value) throws TarantoolValueConverterNotFoundException;
```
akudiyar, line 37: setField is needed for filling the Tuple before sending it to the server. So V will be an Object.

savolgin: Added new method for setting value as Object

PR #4, opened 2020-07-28
Add methods setField and getField by name for TarantoolTuple

savolgin:


src/main/java/io/tarantool/driver/api/tuple/TarantoolTupleImpl.java
```diff
@@ -75,4 +105,63 @@ public void forEach(Consumer<? super TarantoolField> action) {
...
+    public void setField(String fieldName, Object value) {
+        int fieldPosition = getFieldIndexByName(fieldName);
+        if (fieldPosition < 0) {
+            throw new TarantoolSpaceFieldNotFoundException(fieldName);
+        }
+
+        setField(fieldPosition, value);
+    }
+
+    protected int getFieldIndexByName(String fieldName) {
```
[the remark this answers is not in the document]
savolgin: done

src/main/java/io/tarantool/driver/api/tuple/TarantoolTupleImpl.java
```diff
@@ -106,23 +133,27 @@ public int size() {
     }
 
     @Override
-    public void setField(int fieldPosition, Object value) throws MessagePackObjectMapperException {
+    public void setField(int fieldPosition, Object value) {
+        Value messagePackValue = (value == null) ? ValueFactory.newNil() : ((MessagePackObjectMapper) valueMapper).toValue(value);
```
akudiyar, line 137: It is not good to cast value mapper to object mapper, errors are possible. I think we should have a separate constructor taking an object mapper. When a new tuple will be constructed for a request, the constructor with ArrayValue and ValueMapper will not be used.

savolgin: Replaced by separate objectMapper and ValueMapper

savolgin: upd: remove mappers form contructor

src/main/java/io/tarantool/driver/api/tuple/TarantoolFieldImpl.java
```diff
@@ -18,16 +18,14 @@
 public class TarantoolFieldImpl<V extends Value> implements TarantoolField {
 
     private V entity;
-    private MessagePackValueMapper valueMapper;
 
-    TarantoolFieldImpl(V entity, MessagePackValueMapper valueMapper) {
+    TarantoolFieldImpl(V entity) {
```
akudiyar, line 22: Make factory a second parameter and save it to a field, it will be passed from tuple when adding objects to the tuple or may be passed explicitly when constructing fields manually

src/main/java/io/tarantool/driver/api/tuple/TarantoolFieldImpl.java
```diff
@@ -18,16 +18,14 @@
...
         this.entity = entity;
-        this.valueMapper = valueMapper;
     }
 
-    <O> TarantoolFieldImpl(@Nullable O value, MessagePackObjectMapper mapper) {
-        this.entity = value == null ? null : mapper.toValue(value);
-        this.valueMapper = (MessagePackValueMapper) mapper;
+    <O> TarantoolFieldImpl(@Nullable O value) {
+        this.entity = value == null ? null :
+                DefaultMessagePackMapperFactory.getInstance().defaultComplexTypesMapper().toValue(value);
```
akudiyar, line 28: Default factory shouldn't be used implicitly when converting user data

src/main/java/io/tarantool/driver/api/tuple/TarantoolTupleImpl.java
```diff
@@ -19,26 +21,36 @@
...
+    public TarantoolTupleImpl(ArrayValue value) {
+        this(value, null);
+    }
 
     /**
      * Basic constructor.
+     *
      * @param value messagePack entity
-     * @param mapper for converting the entity into the Java objects
+     * @param spaceMetadata tarantool space metadata {@link TarantoolSpaceMetadata}
```
akudiyar, line 43: Leave the factory as a constructor parameter. The previous commit was better

src/main/java/io/tarantool/driver/api/tuple/TarantoolFieldImpl.java
savolgin: done

src/main/java/io/tarantool/driver/api/tuple/TarantoolFieldImpl.java
```diff
@@ -18,16 +18,17 @@
 public class TarantoolFieldImpl<V extends Value> implements TarantoolField {
 
     private V entity;
-    private MessagePackValueMapper valueMapper;
 
-    TarantoolFieldImpl(V entity, MessagePackValueMapper valueMapper) {
+    private DefaultMessagePackMapperFactory mapperFactory;
+
+    TarantoolFieldImpl(V entity, DefaultMessagePackMapperFactory mapperFactory) {
```
akudiyar, line 24: I think we should declare the parameter as interface `M extends MessagePackObjectMapper & MessagePackValueMapper`

akudiyar: done

PR #9, opened 2020-08-11
Implement call request

savolgin:
https://www.notion.so/call-f36f9dde34af4946bab1ea32438ff735

### DIFF
--- a/src/main/java/io/tarantool/driver/TarantoolConnectionImpl.java
+++ b/src/main/java/io/tarantool/driver/TarantoolConnectionImpl.java
@@ -10,8 +10,11 @@ import io.tarantool.driver.metadata.TarantoolMetadataOperations;
 import io.tarantool.driver.metadata.TarantoolSpaceMetadata;
 import io.tarantool.driver.api.space.TarantoolSpace;
 import io.tarantool.driver.api.space.TarantoolSpaceOperations;
+import io.tarantool.driver.protocol.TarantoolProtocolException;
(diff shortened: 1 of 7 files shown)


src/main/java/io/tarantool/driver/protocol/requests/TarantoolCallRequest.java
```diff
@@ -0,0 +1,73 @@
...
+            this.bodyMap.put(TarantoolRequestFieldType.IPROTO_FUNCTION_NAME.getCode(), functionName);
+            return this;
+        }
+
+        /**
+         * Specify function arguments
+         * @param tuple function arguments
+         * @return builder
+         */
+        public TarantoolCallRequest.Builder withArguments(TarantoolTuple tuple) {
```
akudiyar, line 53: I suggest `List<Object>` here,  TarantoolTuple can be an argument type

src/main/java/io/tarantool/driver/TarantoolConnectionImpl.java
```diff
@@ -81,7 +85,32 @@ public TarantoolMetadataOperations metadata() throws TarantoolClientException {
 
     @Override
     public <T> CompletableFuture<T> call(String functionName, Object... arguments) throws TarantoolClientException {
```
akudiyar, line 87: The function variant without converter should return TarantoolResult<T>.

akudiyar: Decided to leave List<T> here, since TarantoolResult needs refactoring

PR #10, opened 2020-08-11
WIP: Cluster discovery connection

savolgin:

PR #14, opened 2020-09-07
Implement multi-connection with service discovery

akudiyar:
- Support for custom address providers
 - HTTP and Binary protocol discovery services and address providers
 - Support for multiple connections per server
 - Cluster and single TarantoolClient implementations

Replaces request #10 (the necessary changes are incorporated)


akudiyar:
TODO: implement reconnection and failover strategies

PR #23, opened 2020-10-18
remove unused deps

isopov:
jsr305 seems to be currently unused (I would suggest use it and it should provided than AFAIU)
logback-core is not needed as compile dep to not come to clients transitively and in tests should come transitively with logback-classic

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -207,11 +207,6 @@
             <version>1.7.30</version>
         </dependency>
         <dependency>
-            <groupId>ch.qos.logback</groupId>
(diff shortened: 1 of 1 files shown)


akudiyar:
logback-core -- ok, seems like extra one
findbugs -- as we discussed earlier, it should replace some used features of a spring dependency. Also, there is a warning in compile-time which indicates missing JSR305 implementation caused by @Nulalble annotation AFAIU.

akudiyar:
Closed in favor of #43

PR #55, opened 2021-02-01
Cursor implementation

vrogach2020:


src/main/java/io/tarantool/driver/api/space/ProxyTarantoolSpaceCursor.java
```diff
@@ -0,0 +1,85 @@
...
+import io.tarantool.driver.exceptions.TarantoolClientException;
+import io.tarantool.driver.exceptions.TarantoolException;
+import io.tarantool.driver.exceptions.TarantoolSpaceOperationException;
+import io.tarantool.driver.mappers.ObjectConverter;
+import io.tarantool.driver.mappers.ValueConverter;
+import org.msgpack.value.ArrayValue;
+
+import java.util.concurrent.ExecutionException;
+
+public class ProxyTarantoolSpaceCursor<T> implements TarantoolCursor<T> {
```
akudiyar, line 15: Docs are missing.
I think it is worth describing here the cursor semantics in regards to the well-known JDBC cursor spec: so that this cursor implementation is a read-only forward-only (or scrollable, we'll discuss it) client-side batching cursor (thus, partially data-sensitive).

src/main/java/io/tarantool/driver/api/space/ProxyTarantoolSpaceCursor.java
```diff
@@ -0,0 +1,85 @@
...
+
+        return result.size() > 0;
+    }
+
+    @Override
+    public T get() throws TarantoolException {
+        if (result.size() <= batchOffset - 1) {
+            throw new TarantoolSpaceOperationException("Batch does not contain elements.");
+        }
+        return result.get((int) batchOffset);
```
akudiyar, line 82: Type narrowing doesn't look good here. Shouldn't the offset be of `int` type? Or maybe we need to introduce the local offset of int type since only a small portion of data is fetched?

src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
```diff
@@ -242,6 +256,22 @@ private boolean checkConverterByTargetType(ValueConverter<? extends Value, ?> co
         objectConvertersByTarget.put(valueClass.getTypeName(), converter);
     }
 
+    @Override
+    public <O, V extends Value> Optional<ObjectConverter<O, V>> getObjectConverter(Class<O> entityClass,
+                                                                                   Class<V> valueClass) {
+        Function<String, Optional<ObjectConverter<O, V>>> getter =
+                typeName -> objectConverters.getOrDefault(typeName, Collections.emptyList()).stream()
+                        .filter(c -> checkObjectConverterByTargetType(c, valueClass))
+                        //FIXME why unchecked cast here ??? -
```
akudiyar, line 265: It's due to type erasure. The type system in Java is not very well developed. We have the same cast in the getValueConverter for the same reason.

akudiyar (changes requested):

Issue #54, opened 2020-12-18
The schema is not received correctly out-of-the-box on Cartridge instances


akudiyar:
# Problem statement

Currently, by default the ProxyTarantoolClient uses `cartridge.get_schema` function for retrieving the schema (see https://github.com/tarantool/cartridge-java/blob/27ef77e3bc1e2ac190d1866701a6f1a3fd7cb33d/src/main/java/io/tarantool/driver/proxy/ProxyOperationsMapping.java#L11). The function result is treated as the result of `ddl.get_schema` API (see https://github.com/tarantool/ddl#input-data-format), but in fact, it returns a YAML string (see https://www.tarantool.io/en/doc/latest/book/cartridge/cartridge_api/modules/cartridge/#get-schema). Also, the README doesn't mention how to set up the TarantoolProxyClient for using a custom function (e.g. ddl.get_schema). This leads to errors when a user tries to use the ProxyTarantoolClient without additional wrappers.

# TODO

- [ ] Support selecting of the metadata converter which will be used for metadata function result conversion
- [ ] Move the current proxy metdata converter to a "DDL schema" converter variant  
- [ ] Parse and use the schema in YAML format returned by `cartridge.get_schema` by a "YAML schema" metadata converter variant
- [ ] Add a section in README describing the Proxy client customization using `ddl.get_schema`

PR #59, opened 2021-02-05
Fix not running integration tests in CI

akudiyar:
Blocked by fixes for #46 (#54)


akudiyar:
These changes are incorporated in #65

PR #65, opened 2021-02-15
Refactor the result type system

### LINKED ISSUE #49: Unexpected tuple format causes MessageTypeCastException in TarantoolResultImpl constructor
Example case:

 - A stored function returns a map `{ rows: ..., metadata: ...}`
 - The stored function result is handled as `TarantoolResult<TarantoolTuple>`
 - Mapper sees the wrong result type and throws the `MessageTypeCastException`

This exception is not readable and requires debugging of the client code.

The possible solution is checking the MsgPack value type and returning a comprehensive error.

### LINKED ISSUE #51: TarantoolCallResultFactory allows concurrent modification of mapper state
Internal cache in TarantoolCallResultMapperFactory produces race conditions when instantiating mappers for the same target type (e.g. TarantoolTuple).

Steps to reproduce:

1. Setup several threads which invoke the `call` method in the ProxyTarantoolClient concurrently with `TarantoolCallResult<TarantoolTuple>` result type but for different entities (space metadata)
2. In some cases the mapper returned by the factory will not correspond to the mapper required by the call result

See https://github.com/tarantool/cartridge-springdata/blob/42f9d1bf0d3c42f4ed8a7a281ba8d357f592dec9/src/main/java/org/springframework/data/tarantool/core/TarantoolTemplate.java#L232 (easy to reproduce with several scheduled threads + API calls)

### LINKED ISSUE #54: The schema is not received correctly out-of-the-box on Cartridge instances

akudiyar:
- Refactored the call result type system, thus allowing to use stored functions returning any result except arrays of tuples
 - Fix parsing ddl.get_schema output (used by default for retrieving metadata in proxy client)
 - Make metadata retrieving and connection establishing retryable (a foundation for further reconnection strategy improvements)
  - Refactored space API, made CRUD operations typed
  - Eliminated the useless StandaloneTarantoolClient
  - Binary cluster discovery endpoint now accepts the address provider and the common Tarantool client config

Fixes #49
Fixes #51
Fixes #54


src/main/java/io/tarantool/driver/metadata/TarantoolMetadata.java
```diff
@@ -1,100 +1,191 @@
...
-        this.mapperFactory = new TarantoolSimpleResultMapperFactory(config.getMessagePackMapper());
+public class TarantoolMetadata implements TarantoolMetadataOperations {
+
+    protected final Map<String, TarantoolSpaceMetadata> spaceMetadataByName = new ConcurrentHashMap<>();
+    protected final Map<Integer, TarantoolSpaceMetadata> spaceMetadataById = new ConcurrentHashMap<>();
+    protected final Map<String, Map<String, TarantoolIndexMetadata>> indexMetadataBySpaceName =
+            new ConcurrentHashMap<>();
+    protected final Map<Integer, Map<String, TarantoolIndexMetadata>> indexMetadataBySpaceId =
+            new ConcurrentHashMap<>();
+    private final Phaser initPhaser = new Phaser(0);
```
vrogach2020, line 27: Please add  a short description what is synced by phaser here and how refresh works

vrogach2020 (approved):

PR #114, opened 2021-07-27
WIP: Added support type cdata double for Tarantool 2.4+

### LINKED ISSUE #104: Add conversion from Double to Tarantool double (cdata)
Since 2.3.x Tarantool supports the `double` type (cdata) for MsgPack values and tuple fields (see https://www.tarantool.io/en/doc/latest/book/box/data_model/). The values of this type may be saved both into `number` or `double` tuple fields. Currently, the Float and Double variable values are converted into MsgPack number, which may lead to losing the information about the source value type, and then, an error occurs while reading the saved value into the Double or Float variable.

We need to implement the correct converter, but it may be necessary to introduce feature toggling for separating the implementations for Tarantool 1.10 and 2.x+.

TODO:

- [ ] Investigate the possibility to save and load `double` values  into `number` fields without losing the type information
- [ ] Check if it is possible to save/load `double` values in 1.10
- [ ] Implement new CdataDoubleValueConverter for 2.x
- [ ] (optional) Implement Tarantool version checking and conditional choice of the converter (a separate task)

wey1and:
Closes #104


vrogach2020 (changes requested):

vrogach2020 (approved):

vrogach2020 (changes requested):
Float.MIN_VALUE is positive, needs fix

akudiyar:
Hi! Thanks for the patch.
Please move "closing #xxx" from title to description.

src/test/java/io/tarantool/driver/mappers/DefaultMessagePackMapperTest.java
```diff
@@ -40,12 +42,9 @@ void getDefaultConverter() throws MessagePackValueMapperException {
         });
         assertEquals(Long.valueOf(111L), mapper.fromValue(ValueFactory.newInteger(111), Long.class));
         assertEquals(Integer.valueOf(111), mapper.fromValue(ValueFactory.newInteger(111L), Integer.class));
-        assertEquals(Float.valueOf(111.0F), mapper.fromValue(ValueFactory.newFloat(111.0F)));
+        assertEquals(Double.MAX_VALUE, mapper.fromValue(ValueFactory.newFloat(Double.MAX_VALUE)));
```
akudiyar, line 45: Seems that Float type conversion check is missing now

src/test/java/io/tarantool/driver/mappers/DefaultMessagePackMapperTest.java
```diff
@@ -40,12 +42,9 @@ void getDefaultConverter() throws MessagePackValueMapperException {
...
         assertEquals(Double.valueOf(Float.MAX_VALUE * 10D),
                 mapper.fromValue(ValueFactory.newFloat(Float.MAX_VALUE * 10D)));
-        assertThrows(ClassCastException.class, () -> {
```
akudiyar, line 46: Why the exception is not thrown here?

akudiyar (changes requested):

vrogach2020 (approved):

src/main/java/io/tarantool/driver/mappers/DefaultFloatValueToIntegerConverter.java
```diff
@@ -19,6 +19,6 @@ public Integer fromValue(FloatValue value) {
     @Override
     public boolean canConvertValue(FloatValue value) {
         double aDouble = value.toDouble();
-        return aDouble >= Integer.MIN_VALUE && aDouble <= Integer.MAX_VALUE;
```
akudiyar: Leave Integer.MIN_VALUE here

src/main/java/io/tarantool/driver/mappers/DefaultFloatValueToShortConverter.java
```diff
@@ -19,6 +19,6 @@ public Short fromValue(FloatValue value) {
     @Override
     public boolean canConvertValue(FloatValue value) {
         double aDouble = value.toDouble();
-        return aDouble >= Short.MIN_VALUE && aDouble <= Short.MAX_VALUE;
```
akudiyar: Leave Short.MIN_VALUE here, it has the right meaning for this operation

akudiyar (changes requested):

src/main/java/io/tarantool/driver/mappers/DefaultFloatValueToIntegerConverter.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public Integer fromValue(FloatValue value) {
+        return value.toInt();
+    }
+
+    @Override
+    public boolean canConvertValue(FloatValue value) {
+        double aDouble = value.toDouble();
+        return aDouble >= Integer.MIN_VALUE && aDouble <= Integer.MAX_VALUE;
```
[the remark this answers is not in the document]
akudiyar: Discussed: the converters stack is actually Short -> Integer -> Long, so this expression is right

src/main/java/io/tarantool/driver/mappers/DefaultFloatConverter.java
```diff
@@ -24,6 +24,7 @@ public Float fromValue(FloatValue value) {
 
     @Override
     public boolean canConvertValue(FloatValue value) {
-        return value.toDouble() <= Float.MAX_VALUE;
+        double aDouble = value.toDouble();
+        return aDouble >= -Float.MAX_VALUE && aDouble <= Float.MAX_VALUE;
```
akudiyar, line 28: Seems that this expression is too simple. You didn't take into account the lowest rational (FLOAT.MIN_VALUE) -- the right solution is taking abs() and checking the value within the type boundaries

akudiyar (changes requested):

akudiyar:
The code seems good now, but I think we miss some tests for `tuple.getObject("field", aClass)`. We need to check that the default stack of converters in DefaultMessagePackMapper works as expected in the following cases:

1. User tries to read a `Short` value (the returned value must have the type `Short`)
2. User tries to read an `Integer` value greater than `Short` type bounds (the returned value must have the type `Integer`)
3. User tries to read an `Integer` value within the `Short` type bounds (the returned value must have the type `Integer`)
4. User tries to read a `Short` value but the value is out of the `Short` type bounds (NoConverterAvailableException is expected)
5. User tries to read an `Integer` value (the returned value must have the type `Integer`)
6. User tries to read a `Long` value greater than `Integer` type bounds (the returned value must have the type `Long`)
7. User tries to read a `Long` value within the `Integer` type bounds (the returned value must have the type `Long`)
8. User tries to read an `Integer` value but the value is out of the `Integer` type bounds (NoConverterAvailableException is expected)

Issue #110, opened 2021-07-16
TarantoolClient factory and builders


akudiyar:
Currently, the client must be instantiated with not so convenient way:

```java
    TarantoolClientConfig config = TarantoolClientConfig.builder()
            // use the value of cluster_cookie parameter in the init.lua file in your Cartridge application
            .withCredentials(new SimpleTarantoolCredentials("admin", "secret-cluster-cookie"))
            .build();

    ClusterTarantoolTupleClient clusterClient = new ClusterTarantoolTupleClient(config, ROUTER_HOST, ROUTER_PORT);
    TarantoolClient client = new ProxyTarantoolTupleClient(clusterClient);
```

More lines are needed for creating a retrying setup:

```java
     client = new RetryingTarantoolTupleClient(client,
            TarantoolRequestRetryPolicies.byNumberOfAttempts(
            retries, e -> e.getMessage().contains("Unsuccessful attempt")));
```

Further, the number of client implementations and their options will grow and the configuring code will become an unreadable bulk.

There is a known solution for such a problem -- introducing a client factory and builders for different types of clients. How it may look like:

```java
    TarantoolClientFactory factory = new TarantoolClientFactory(...);
    client = factory.retryingClient()
         .withCredentials(new SimpleTarantoolCredentials("admin", "secret-cluster-cookie"))
         .withAddress(ROUTER_HOST, ROUTER_PORT)
         .withRetryPolicy(TarantoolRequestRetryPolicies.byNumberOfAttempts(retries, e -> e.getMessage().contains("Unsuccessful attempt")))
         .build();
```

PR #125, opened 2021-09-06
WIP: Add client builders and factory

wey1and:
Closes https://github.com/tarantool/cartridge-java/issues/110

### DIFF
--- a/src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilder.java
+++ b/src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilder.java
@@ -1,0 +1,17 @@
+package io.tarantool.driver.clientbuilder;
+
+import io.tarantool.driver.ProxyTarantoolTupleClient;
+import io.tarantool.driver.api.TarantoolClient;
+import io.tarantool.driver.api.TarantoolResult;
+import io.tarantool.driver.api.tuple.TarantoolTuple;
+import io.tarantool.driver.proxy.ProxyOperationsMappingConfig;
+
+public interface TarantoolProxyClientBuilder
+        extends TarantoolDecoratedClientBuilder
+        <ProxyTarantoolTupleClient, TarantoolProxyClientBuilder,
+                TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>>> {
+
+    TarantoolProxyClientBuilder INSTANCE = new TarantoolProxyClientBuilderImpl();
+
+    TarantoolProxyClientBuilder withMappingConfig(ProxyOperationsMappingConfig mappingConfig);
+}
--- a/src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilderImpl.java
+++ b/src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilderImpl.java
@@ -1,0 +1,44 @@
+package io.tarantool.driver.clientbuilder;
+
+import io.tarantool.driver.ClusterTarantoolTupleClient;
+import io.tarantool.driver.ProxyTarantoolTupleClient;
+import io.tarantool.driver.api.TarantoolClient;
+import io.tarantool.driver.api.TarantoolResult;
+import io.tarantool.driver.api.tuple.TarantoolTuple;
+import io.tarantool.driver.proxy.ProxyOperationsMappingConfig;
+import io.tarantool.driver.utils.Assert;
+
+class TarantoolProxyClientBuilderImpl
+        extends AbstractTarantoolDecoratedClientBuilder
+        <ProxyTarantoolTupleClient, TarantoolProxyClientBuilder,
+                TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>>>
+        implements TarantoolProxyClientBuilder {
+
+    private ProxyOperationsMappingConfig mappingConfig;
+
+    @Override
+    public ProxyTarantoolTupleClient build() {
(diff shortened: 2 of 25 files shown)


src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilder.java
```diff
@@ -0,0 +1,17 @@
...
+import io.tarantool.driver.ProxyTarantoolTupleClient;
+import io.tarantool.driver.api.TarantoolClient;
+import io.tarantool.driver.api.TarantoolResult;
+import io.tarantool.driver.api.tuple.TarantoolTuple;
+import io.tarantool.driver.proxy.ProxyOperationsMappingConfig;
+
+public interface TarantoolProxyClientBuilder
+        extends TarantoolDecoratedClientBuilder
+        <ProxyTarantoolTupleClient, TarantoolProxyClientBuilder,
+                TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>>> {
```
ArtDu, line 12: Maybe that's more readable? (for all big generics)
```suggestion
<ProxyTarantoolTupleClient,
 TarantoolProxyClientBuilder,
 TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>>> {
```

src/main/java/io/tarantool/driver/clientbuilder/TarantoolProxyClientBuilderImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public TarantoolProxyClientBuilderImpl() {
+        initBuilder();
+    }
+
+    @Override
+    protected void initBuilder() {
+        super.instance = this;
+    }
+}
```
ArtDu, line 44: In all implementations, we have the same code, maybe there is an opportunity to take him higher to deduplicate

ArtDu (commented):
Thanks for the patchset!
I just reviewed only the Builders commit.
See my comments inline and I have general comments regarding the whole patchset. I know that is the only WIP, but please add a description or explanation of changes to a commit message. This will make it easier to review.

src/main/java/io/tarantool/driver/clientbuilder/TarantoolRetryingClientBuilderImpl.java
```diff
@@ -0,0 +1,46 @@
...
+            return new RetryingTarantoolTupleClient(super.getDecoratedClient(), this.retryPolicyFactory);
+        }
+        return new RetryingTarantoolTupleClient(
+                super.getDecoratedClient(), this.retryPolicyFactory, this.executor);
+    }
+
+    @Override
+    protected void initBuilder() {
+        super.instance = this;
+    }
```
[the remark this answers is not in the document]
akudiyar: I also do not get this pattern. And seems that it is very bad: we modify some global state (a singleton instance in a class) with some concrete implementation with its own internal state. It breaks several design principles at once.

Consider using normal class instances created at their point of usage (in the Tarantool Client Factory).

src/main/java/io/tarantool/driver/clientfactory/AbstractTarantoolClientBuilderDecorator.java
```diff
@@ -0,0 +1,77 @@
...
+import io.tarantool.driver.TarantoolServerAddress;
+import io.tarantool.driver.auth.TarantoolCredentials;
+import io.tarantool.driver.clientbuilder.TarantoolClientBuilder;
+import io.tarantool.driver.clientbuilder.TarantoolClusterClientBuilder;
+import io.tarantool.driver.proxy.ProxyOperationsMappingConfig;
+import io.tarantool.driver.retry.RequestRetryPolicyFactory;
+
+import java.util.concurrent.Executor;
+
+public abstract class AbstractTarantoolClientBuilderDecorator<T, B extends TarantoolClientBuilder<T, B>>
```
akudiyar, line 15: What's the meaning of this class, what problem does it solve?

src/main/java/io/tarantool/driver/clientfactory/TarantoolClientBuilders.java
```diff
@@ -0,0 +1,5 @@
+package io.tarantool.driver.clientfactory;
+
+public interface TarantoolClientBuilders {
```
akudiyar, line 3: Some leftover?

src/main/java/io/tarantool/driver/clientfactory/TarantoolClientFactory.java
```diff
@@ -0,0 +1,12 @@
+package io.tarantool.driver.clientfactory;
+
+public interface TarantoolClientFactory {
+
+    TarantoolClientFactory INSTANCE = new TarantoolClientFactoryImpl();
```
akudiyar, line 5: Singleton factory instances are needed if they keep some state. In the other cases, static factory methods are enough.

src/main/java/io/tarantool/driver/clientfactory/TarantoolRetyingClientBuilderDecoratorImpl.java
```diff
@@ -0,0 +1,49 @@
+package io.tarantool.driver.clientfactory;
```
akudiyar, line 1: Why the factory and builder classes are not in the `api` package or in the root one, where the `TarantoolClient` resides?

The current driver structure implies the separation of classes into packages by domain, not by patterns they implement.

akudiyar (changes requested):

PR #127, opened 2021-09-08
use maven wrapper in github actions

isopov:
There is maven wrapper in project but for some reason it is not used on github actions. For better similarity between local development and continuous integration it seems better to use it everywhere.

### DIFF
--- a/.github/workflows/ubuntu-master.yml
+++ b/.github/workflows/ubuntu-master.yml
@@ -20,9 +20,10 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up JDK 1.8
-      uses: actions/setup-java@v1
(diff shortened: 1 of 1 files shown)


akudiyar:
I'm not sure about this change. `mvn` is being installed in CI using the `setup-java` action, so I see no reason for invoking maven wrapper for searching or installing `mvn` if it apparently exists. For local (or new) environments it does help, but in CI is not necessary.

isopov:
One of the reasons to use wrapper is to ensure that build on every machine will use one and the same maven version. For this reason I suggest to use it in CI

isopov:
Also wrapper will not be downloaded every built - it is stored inside `~/.m2/wrapper/` and this folder is already cached in github actions.

PR #134, opened 2021-09-15
DRAFT (comments for client builder wizard)

vrogach2020:
Fixed class naming and added testcase to implement (retrying cluster client)


akudiyar:
Let me elaborate a bit about the wizard implementation idea.

Although the client creation may be represented in a wizard-like way, and we already have some patterns which seem to fit into this, there are some downsides of using such an approach from a long-term perspective:

1. Wizard-like interfaces are not as universal as builder-like interfaces. The builder-like interface is easier to refactor if we need to add a new option or a new Client type, which may form either only some specific combinations with other Client types or be applicable to all known types, which leads to a combinatoric complexity of changes.
2. The idea of different client types is based on the separation of the wrapping code which changes the client behavior (adds request proxying with different implementations, request repetition, request logging, etc) and the connection state and internal transport mechanisms, which is achieved by using the well-known "Decorator" pattern. This pattern also allows performing an idempotent combination of the client types to form any allowed hybrid types of clients like "Cluster + Retrying", "Cluster + CRUD + Retrying", "Cluster + CustomProxy + Retrying" ad so on. The builder-like interface does not require much effort for supporting the combination logic: we may achieve this by creating an abstract builder implementation that supports all known types of clients via factory methods and disclose only the supported methods in the derived builder classes. As for the wizard-like interface, we'll have to provide the "step interface" variants for each possible combination, thus receiving more code and some weird interface names. There probably may be better uses of both ideas, so we may discuss them.
3. Wizard-like interfaces are harder to reuse than builder-like if we want to provide the means of reusing the configuration. Consider a use case, when a user configures a CustomProxy client as a bean and then wraps it into a retrying interface for some critical idempotent methods. In the case of a wizard, the user will receive either a client or a wizard step interface, but with the builder-like implementation, he/she will be able to reuse a configured builder by adding any configuration to it via the allowed factory methods. A "step" interface will just look weird for reusing.
4. The wizard interface requires the user to proceed over all the steps, manually specifying "defaults" even if the user doesn't want to customize them, although the user wants to write less code. The builder-like interface allows writing only the necessary lines with customizations.

vrogach2020:
Discussed. Will go to builder implementation with idempotent methods.

Issue #39, opened 2020-11-13
Update by field name is not working


akudiyar:
When trying to update a tuple by field name, the following error occurs:
```
Illegal parameters, field id must be a number or a string
```

The reason is missing field name serialization in `TarantoolUpdateOperation`:

```java
    @Override
    public Value toMessagePackValue(MessagePackObjectMapper mapper) {
        return mapper.toValue(
                Arrays.asList(getOperationType().toString(), getFieldIndex(), getValue()));
    }
```

wey1and:
This PR closes this issue
https://github.com/tarantool/cartridge-java/pull/140

PR #140, opened 2021-10-14
Fix upsert operations serialization

### LINKED ISSUE #139: Upsert in proxy client serializes to messagepack incorrectly
When field in upsert condition is defined by name (not by index) then it becomes null after serializaion.

vrogach2020:
This PR fixes serialization of TupleUpdateOperation and TupleSpliceOperation when field name is set. Closes #139


wey1and:
Closes #39

Issue #144, opened 2021-10-24
Writing byte[] values into varbinary fields fails


akudiyar:
Having the following space structure:
```lua
    local test_space_with_byte_array = box.schema.space.create(
             'test_space_with_byte_array', { if_not_exists = true })

    test_space_with_byte_array:format({
        { name = 'id', type = 'unsigned' },
        { name = 'bucket_id', type = 'unsigned' },
        { name = 'value', type = 'varbinary' },
    });
```
and trying to write a byte value into it:
```java
client.space(SPACE)
                .insert(tupleFactory.create(1, null, new byte[]{1, 2, 3, 4})).get();
```
it fails with the following error:
```
java.util.concurrent.ExecutionException: io.tarantool.driver.exceptions.TarantoolInternalException: InnerErrorMessage:
str: Insert: Failed to insert: Call: Failed for d1bf9bff-b57d-4f8b-804c-8f34e24fdb14: Function returned an error: {"code":32,"base_type":"ClientError","type":"ClientError","message":"Tuple field 3 (value) type does not match one required by operation: expected varbinary, got string","trace":[{"file":"[C]","line":4294967295}]}
stack traceback:
	/app/.rocks/share/tarantool/crud/common/call.lua:163: in function 'rw_single'
	/app/.rocks/share/tarantool/crud/insert.lua:71: in function </app/.rocks/share/tarantool/crud/insert.lua:52>
	[C]: at 0x004fdd50
line: 76
class_name: Insert
err: /app/.rocks/share/tarantool/crud/insert.lua
```

Expected behavior: byte arrays are correctly encoded and saved into `varbinary` fields.

Related to https://github.com/tarantool/tarantool/issues/1629

ArtDu:
### space operations (IPROTO_SELECT, IPROTO_INSERT and etc.)

We have correctly varbinary work with space iproto methods without touching the lua world(i.e. not IPROTO_EVAL/CALL) (you can see the correct work here https://github.com/tarantool/cartridge-java/pull/228).

### lua world operations (IPROTO_CALL, IPROTO_EVAL)
In the lua world, we accept a string because there is no easy way to pass MP_BIN to lua without `ffi` https://github.com/tarantool/tarantool/issues/1629 . That is the reason why crud functions don't work correctly. 

> Possible solution: implement encoding procedure as suggested in [tarantool/tarantool#1629 (comment)](https://github.com/tarantool/tarantool/issues/1629#issuecomment-705097392)

[...]

PR #146, opened 2021-10-24
Add integration tests for byte array converter

akudiyar:
Affects #144


ArtDu:
The tests were added in https://github.com/tarantool/cartridge-java/pull/228. [Test](https://github.com/tarantool/cartridge-java/blob/0676de2056e65e4dda7f0cf599881bea8b3c02f3/src/test/java/io/tarantool/driver/integration/ConvertersWithProxyClientIT.java#L68-L87) with varbinary and crud has been disabled, because we can't use varbinary with crud until https://github.com/tarantool/tarantool/issues/1629 fixes it

PR #147, opened 2021-10-25
Added methods in client builder

wey1and:
A new method has been added to simplify messagepack mapper configuration.

The user no longer needs to know about the default mapper and look for it. Now it is provided with a lambda with an interface for a simple mapper configuration. There is also a method for advanced users to allow them to pass their own mapper.

Also added method for specifying TarantoolClientConfig in TarantoolClient builder


akudiyar:
Let's fix the comments and merge this one too, why does it stay still?

vrogach2020 (approved):

akudiyar:
@wey1and please add tests for the new default mapper and config options.

akudiyar (approved):

PR #151, opened 2021-11-19
Fix problems with TarantoolSpace.update/upsert

### LINKED ISSUE #63: incorrect behavior when trying to update a field with fieldName
Data loss problem. An error is expected in assert, but it updates the last field if the fieldName that is specified in update does not exist:
```java
updateResult = testSpace.update(conditions, TupleOperations.add("asd", 100000)).get();
assertEquals(101968, updateResult.get(0).getInteger(4));
```
This line changes fieldId from None to -1:
https://github.com/tarantool/cartridge-java/blob/e6f00e0e07fecebe6a9af030dfd342ef2c27100d/src/main/java/io/tarantool/driver/api/space/TarantoolSpace.java#L235

ArtDu:
This pr resolves two related problems:

1. We use update / upsert from TarantoolSpace, which updates the space via iproto. To specify the field to update, use fieldName. Previously, if we made a mistake in the name, no error occurred and a request was sent to tarantool where fieldIndex = -1, and this is correct for tarantool because it uses reverse index, so -1 is the last element in the Tuple. For now, we'll throw a TarantoolSpaceFieldNotFoundException if the specified fieldName doesn't exist.

2. The user creates operations by fieldName, which are subsequently used by the space.update/upsert functions. After using these functions, a field index field appears in operations that were missing. To remedy this situation, we make TupleOperation non-mutable. And every time we need to change the fields of these operations, for example, in the space.update/upsert functions, then we clone these operations.

Closes #63

### DIFF
--- a/src/main/java/io/tarantool/driver/core/space/TarantoolSpace.java
+++ b/src/main/java/io/tarantool/driver/core/space/TarantoolSpace.java
@@ -257,14 +257,15 @@ public abstract class TarantoolSpace<T extends Packable, R extends Collection<T>
 
     private List<TupleOperation> fillFieldIndexFromMetadata(TupleOperations operations) {
         return operations.asList().stream()
-                .peek(operation -> {
+                .map(operation -> {
                     if (operation.getFieldIndex() == null) {
                         String fieldName = operation.getFieldName();
                         int fieldMetadataIndex = this.spaceMetadata.getFieldByName(fieldName)
                                 .orElseThrow(() -> new TarantoolSpaceFieldNotFoundException(fieldName))
                                 .getFieldPosition();
-                        operation.setFieldIndex(fieldMetadataIndex);
+                        return operation.cloneWithIndex(fieldMetadataIndex);
                     }
+                    return operation;
                 }).collect(Collectors.toList());
     }
 }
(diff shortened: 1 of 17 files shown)


wey1and (dismissed):

src/main/java/io/tarantool/driver/core/space/TarantoolSpace.java
```diff
@@ -250,4 +254,17 @@ public String toString() {
...
+        return operations.asList().stream()
+                .peek(operation -> {
+                    if (operation.getFieldIndex() == null) {
+                        String fieldName = operation.getFieldName();
+                        int fieldMetadataIndex = this.spaceMetadata.getFieldByName(fieldName)
+                                .orElseThrow(() -> new TarantoolSpaceFieldNotFoundException(fieldName))
+                                .getFieldPosition();
+                        operation.setFieldIndex(fieldMetadataIndex);
+                    }
+                }).collect(Collectors.toList());
```
ArtDu, line 268: it doesn't seem to work even if we do a new ArrayList (operations), since this is a nested structure - deepcopy needed. And we can do this if we add a clone to the class that implements the `TupleOperation` interface. Therefore, you can still think about whether we need to copy operations or still we will change user operations.

akudiyar:
Add a more detailed description or a link to an issue where the current and expected behaviors are described, please.

ArtDu:
> Add a more detailed description or a link to an issue where the current and expected behaviors are described, please.

The link to the ticket already is in the description, If there is not enough description in the ticket, then I will add it

akudiyar:
@ArtDu OK, thanks for pointing this out. Would you mind adding the description of the changes then?

ArtDu:
> @ArtDu OK, thanks for pointing this out. Would you mind adding the description of the changes then?

Sure, I have now updated the description

src/main/java/io/tarantool/driver/core/space/TarantoolSpace.java
ArtDu: Fixed by copying of Operations

vrogach2020 (approved):

Issue #149, opened 2021-11-08
Check that cartridge-java does not create connections above limit


vrogach2020:
We observe a strange behavior of a cluster. 
It looks like java client creates lots of connections (10 times more than expected).
Cluster config:
~ 48 routers, 200 storages
~ 20 clients, 192 connections per client

Expected number of IPROTO connections in cluster ~ 5000.
Probably after some network errors number of connections increased up to 55k and remain stable for about 24 hours. 

What we can do here:

- [ ] add connection count metrics to java client
- [ ] cover connection limit by unit / integration tests. Consider retrying and non-retrying configurations

vrogach2020:
Also we should check how Netty closes sockets and if it is handled in proper way.
See https://stackoverflow.com/questions/28812858/netty-server-does-not-close-release-socket

wey1and:
During discussions and debugging, the following problems were identified:
1) We probably have a race condition when accessing the ConnectionRegister - we need to add thread synchronization in the `AbstractTarantoolConnectionManager` class.
2) It is necessary to add meaningful metrics inside the connection mechanism.
3) It is necessary to add a connection strategy, where the user can specify the following parameters: when and under what circumstances it is necessary to reconnect.
4) Add more logging to `AbstractTarantoolConnectionManager`.
5) We need to think about configuring the number of `nioEventLoops` threads. Why is their number growing and do we need to stop it?

PR #156, opened 2021-11-26
WIP: Changed connection logic

wey1and:
Fixes https://github.com/tarantool/cartridge-java/issues/149

### DIFF
--- a/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
+++ b/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
@@ -65,7 +65,7 @@ public abstract class AbstractTarantoolClient<T extends Packable, R extends Coll
     private final Bootstrap bootstrap;
     private final TarantoolConnectionFactory connectionFactory;
     private final TarantoolConnectionListeners listeners;
-    private final AtomicReference<TarantoolConnectionManager> connectionManagerHolder = new AtomicReference<>();
+    private TarantoolConnectionManager connectionManager;
     private final AtomicReference<TarantoolMetadata> metadataHolder = new AtomicReference<>();
     private final DefaultResultMapperFactoryFactory mapperFactoryFactory;
     private final SpacesMetadataProvider metadataProvider;
@@ -138,15 +138,14 @@ public abstract class AbstractTarantoolClient<T extends Packable, R extends Coll
                                                                     TarantoolConnectionListeners listeners);
 
     private TarantoolConnectionManager connectionManager() {
-        if (this.connectionManagerHolder.get() == null) {
+        if (this.connectionManager == null) {
             synchronized (this) {
-                if (this.connectionManagerHolder.get() == null) {
-                    this.connectionManagerHolder.compareAndSet(null,
-                            connectionManager(config, connectionFactory, listeners));
+                if (this.connectionManager == null) {
+                    this.connectionManager = connectionManager(config, connectionFactory, listeners);
                 }
             }
         }
-        return connectionManagerHolder.get();
+        return connectionManager;
     }
 
     @Override
--- a/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
+++ b/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
@@ -139,7 +139,12 @@ public abstract class AbstractTarantoolClient<T extends Packable, R extends Coll
 
     private TarantoolConnectionManager connectionManager() {
         if (this.connectionManagerHolder.get() == null) {
-            this.connectionManagerHolder.compareAndSet(null, connectionManager(config, connectionFactory, listeners));
+            synchronized (this) {
+                if (this.connectionManagerHolder.get() == null) {
+                    this.connectionManagerHolder.compareAndSet(null,
+                            connectionManager(config, connectionFactory, listeners));
+                }
+            }
         }
         return connectionManagerHolder.get();
     }
--- a/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
+++ b/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
@@ -27,7 +27,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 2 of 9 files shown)


src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
```diff
@@ -139,7 +139,12 @@ protected abstract TarantoolConnectionManager connectionManager(TarantoolClientC
 
     private TarantoolConnectionManager connectionManager() {
         if (this.connectionManagerHolder.get() == null) {
-            this.connectionManagerHolder.compareAndSet(null, connectionManager(config, connectionFactory, listeners));
+            synchronized (this) {
+                if (this.connectionManagerHolder.get() == null) {
+                    this.connectionManagerHolder.compareAndSet(null,
```
vrogach2020, line 144: We do not need atomic reference here since there is double checked lock.

vrogach2020: UPD think about making connectionManagerHolder volatile

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -96,78 +86,68 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
                     ex = ex.getCause();
                 }
                 if (ex instanceof NoAvailableConnectionsException) {
-                    connectionMode.set(ConnectionMode.FULL);
+                    //todo:
```
vrogach2020, line 89: Write a comment of what you are planing TO DO here please

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -96,78 +86,68 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
...
                 }
                 throw new TarantoolConnectionException(ex);
             }
             return connection;
         });
     }
 
+    public void connect() {
+        establishConnections().thenAccept(registry -> {
+            connectionRegistry.set(registry);
```
vrogach2020, line 99: Why is registry reset on every execution ?

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -41,27 +45,10 @@
     private final AtomicReference<Map<TarantoolServerAddress, List<TarantoolConnection>>> connectionRegistry =
```
vrogach2020, line 45: Do we really need atomics here for registry and strategy ? And why ?

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -96,78 +86,68 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
...
-                    connectionMode.set(ConnectionMode.FULL);
+                    //todo:
                 }
                 throw new TarantoolConnectionException(ex);
             }
             return connection;
         });
     }
 
+    public void connect() {
```
vrogach2020, line 97: This code has a very bad readabilty.

We have methods:
getConnection()
getConnectionInternal()
initConnections()
initReconnectService()
connect()
establishConnections()
getConnections()
establishConnectionsToEndPoint()


All these method names look similar. 
The methods are doing lots of  things - creating objects, locking semaphores. etc. and it is not clear from their names at all.

I think this code needs a heavy refactoring:
 - there should be more descriptive method names and comments
 - try to follow the single responsibilty principle

vrogach2020 (changes requested):

akudiyar:
Add a description or a link to an issue with RFC to this task.

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
akudiyar: Why is this method needed? It will require each relying library to create and maintain connections by itself. This method is not compatible with lazy init and context manager patterns (try-with-resources).

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -96,78 +85,68 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
...
-            // In this case the calling code may perform the request again.
-            result = new CompletableFuture<>();
-            try {
-                result.complete(connectionSelectStrategy.get().next());
-            } catch (Throwable t) {
-                result.completeExceptionally(t);
-            }
+        CompletableFuture<TarantoolConnection> result = new CompletableFuture<>();
+
+        if (connectionRegistry.isEmpty()) {
```
akudiyar, line 110: isEmpty() is not concurrent safe, as the HashMap itself. You will get multiple initConnections() calls. Either use some other Map implementation, or the double-checked locking pattern one more time.

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -96,78 +85,68 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
...
         CompletableFuture<Map<TarantoolServerAddress, List<TarantoolConnection>>> result = new CompletableFuture<>();
         try {
-            List<CompletableFuture<Map.Entry<TarantoolServerAddress, List<TarantoolConnection>>>> endpointConnections =
-                    getConnections();
-            result = CompletableFuture
-                    .allOf(endpointConnections.toArray(new CompletableFuture[0]))
-                    .thenApply(v -> endpointConnections.parallelStream()
-                            .map(CompletableFuture::join)
-                            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
+            result = CompletableFuture.supplyAsync(() -> getConnections().parallelStream()
```
akudiyar, line 147: Always use a specific executor and not the built-in one in such places. parallelStream() invokes a common executor for every such call, so the driver code may compete for resources with some other code.

akudiyar (changes requested):
Let's discuss the changes verbally in the near time.
The most important things that are missing here:
1. Description of every change (a detailed RFC). It is not a simple piece of logic, and every reviewer must get the idea first, for not doing reverse engineering and in-head-compiling each time.
2. Separation of fixes from the reconnection policy implementation. It definitely needs an RFC, since it's also not a simple change.
3. Tests for the scenarios to be fixed and new tests for the new behavior. Ideally, some load test suite, that will be added along with this PR and be run periodically forever.

PR #158, opened 2021-12-02
Fixed connection manager logic

wey1and:
Fixed concurrent access to the creation of a connection manager and its internal logic for establishing connections in the event of a partial or complete failure.
Fixes https://github.com/tarantool/cartridge-java/issues/149

### DIFF
--- a/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
+++ b/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
@@ -200,7 +200,7 @@ public abstract class AbstractTarantoolConnectionManager implements TarantoolCon
                 for (TarantoolConnection aliveConnection : aliveConnections) {
                     if (count-- > 0) {
                         try {
-                            logger.info("Closing connection to {}, connections size greater than {}",
+                            logger.info("Closing connection to {}, connections size is greater than {}",
                                     aliveConnection.getRemoteAddress(), config.getConnections());
                             aliveConnection.close();
                         } catch (Exception e) {
@@ -215,10 +215,8 @@ public abstract class AbstractTarantoolConnectionManager implements TarantoolCon
                         new AbstractMap.SimpleEntry<>(serverAddress, aliveConnections)));
             }
         }
-
         return endpointConnections;
     }
-
 
     private List<TarantoolConnection> getAliveConnections(TarantoolServerAddress serverAddress) {
         List<TarantoolConnection> connections = connectionRegistry.get(serverAddress);
--- a/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
+++ b/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
@@ -38,8 +38,7 @@ public abstract class AbstractTarantoolConnectionManager implements TarantoolCon
     private final TarantoolConnectionFactory connectionFactory;
     private final ConnectionSelectionStrategyFactory selectStrategyFactory;
     private final TarantoolConnectionListeners connectionListeners;
-    private final AtomicReference<Map<TarantoolServerAddress, List<TarantoolConnection>>> connectionRegistry =
-            new AtomicReference<>(new HashMap<>());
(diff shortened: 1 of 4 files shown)


src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -207,12 +215,17 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
                         new AbstractMap.SimpleEntry<>(serverAddress, aliveConnections)));
             }
         }
+
         return endpointConnections;
     }
 
+
```
akudiyar, line 222: Extra newline

src/main/java/io/tarantool/driver/core/connection/ConnectionMode.java
```diff
@@ -18,5 +18,7 @@
     /**
      * Init sequence completed, requests will not block
      */
-    OFF;
+    OFF,
+
+    IN_PROGRESS;
```
akudiyar, line 23: Documentation is required

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -207,12 +215,17 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
                         new AbstractMap.SimpleEntry<>(serverAddress, aliveConnections)));
             }
         }
+
```
akudiyar, line 218: Extra newline

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -194,6 +200,8 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
                 for (TarantoolConnection aliveConnection : aliveConnections) {
                     if (count-- > 0) {
                         try {
+                            logger.info("Closing connection to {}, connections size greater than {}",
```
akudiyar, line 203: is greater

akudiyar (changes requested):

akudiyar:
I'd like to add the more detailed description of what's changed:

Previously, when a connection to a host is being failed for some time under constant load, several of the request threads which tried to acquire a connection may turn on the connection mode simultaneously (how many -- depends on the request rate and the timeout to wait until a connection fails). This was happening because there were only two connection states -- "ON" (represented by FULL/PARTIAL) and "OFF", and this led to a situation when one thread finished connecting, a second is waiting until the connection process finishes, and a third starts connecting immediately as it catches the NoAvailableConnectionsException and sees the connection state as "OFF".

In this patch, a new connection state is added -- IN_PROGRESS, and such race condition is not more possible since only one thread actually can set the state to IN_PROGRESS, and only it can switch it back when the connection process is finished (succeed or failed).

vrogach2020 (approved):

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -211,8 +220,11 @@ public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
     }
 
     private List<TarantoolConnection> getAliveConnections(TarantoolServerAddress serverAddress) {
-        List<TarantoolConnection> connections = connectionRegistry.get()
-                .getOrDefault(serverAddress, Collections.emptyList());
+        List<TarantoolConnection> connections = connectionRegistry.get(serverAddress);
```
akudiyar, line 223: I didn't get the reason of this change. Some debugging leftovers maybe? Let's change it back.

akudiyar (changes requested):

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
vrogach2020: +1 the previous version looks cuter

akudiyar (approved):

vrogach2020 (approved):

PR #160, opened 2022-01-10
Add method for reconnection

wey1and:
Closes https://github.com/tarantool/cartridge-java/issues/115
Added a method to switch the connection mode to `PARTIAL` to start the process of establishing the lacking connections.

### DIFF
--- a/src/test/java/io/tarantool/driver/integration/ReconnectIT.java
+++ b/src/test/java/io/tarantool/driver/integration/ReconnectIT.java
@@ -19,8 +19,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 @Testcontainers
 public class ReconnectIT extends SharedCartridgeContainer {
 
-    public static String USER_NAME;
-    public static String PASSWORD;
+    private static String USER_NAME;
+    private static String PASSWORD;
 
     @BeforeAll
     public static void setUp() throws TimeoutException {
@@ -45,16 +45,18 @@ public class ReconnectIT extends SharedCartridgeContainer {
                                 new TarantoolServerAddress(container.getRouterHost(), container.getMappedPort(3312))
                         )
                         .withCredentials(USER_NAME, PASSWORD)
+                        .withConnections(10)
                         .build();
 
+        // getting all routers uuids
         final Set<String> routerUuids = getInstancesUuids(client);
 
         // stop routers
         container.execInContainer("cartridge", "stop", "--run-dir=/tmp/run", "router");
         container.execInContainer("cartridge", "stop", "--run-dir=/tmp/run", "second-router");
 
-        assertEquals(client.eval("return box.info().uuid").join().get(0),
-                client.eval("return box.info().uuid").join().get(0));
+        // check that there is only one instance left
+        assertEquals(getInstanceUuid(client), getInstanceUuid(client));
 
         // start routers
         container.execInContainer("cartridge", "start", "--run-dir=/tmp/run", "--data-dir=/tmp/data", "-d");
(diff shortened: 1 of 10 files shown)


src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.TimeoutException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Testcontainers
+public class ReconnectIT extends SharedCartridgeContainer {
+
+    public static String USER_NAME;
```
vrogach2020, line 22: why static ? and public ?

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -0,0 +1,86 @@
...
+                        .withCredentials(USER_NAME, PASSWORD)
+                        .build();
+
+        final Set<String> routerUuids = getInstancesUuids(client);
+
+        // stop routers
+        container.execInContainer("cartridge", "stop", "--run-dir=/tmp/run", "router");
+        container.execInContainer("cartridge", "stop", "--run-dir=/tmp/run", "second-router");
+
+        assertEquals(client.eval("return box.info().uuid").join().get(0),
```
vrogach2020, line 56: Flacky check. You can create a client with a connection to a single (switched off router) to check that it is down. Something like
```
bool isDown(int port) {

  try { 
    TarantoolClientFactory.createClient()
                        .withAddresses(
                                new TarantoolServerAddress(container.getRouterHost(), container.getMappedPort(port)),
                        )
                        .withCredentials(USER_NAME, PASSWORD)
                        .build();
   client.getVersion().get();
   return true;
  } catch( TarantoolConnectionException ex ) {
    ...
  }
  return false;
}
```

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -0,0 +1,86 @@
...
+
+        // start routers
+        container.execInContainer("cartridge", "start", "--run-dir=/tmp/run", "--data-dir=/tmp/data", "-d");
+
+        client.establishLackingConnections();
+        Thread.sleep(3000);
+
+        final Set<String> uuidsAfterReconnect = getInstancesUuids(client);
+
+        assertEquals(routerUuids.size(), uuidsAfterReconnect.size());
```
vrogach2020, line 67: I think the more robust check is to use
```
localhost:3301> box.stat.net().CONNECTIONS
---
- current: 3
  rps: 0
  total: 4
...
```

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private Set<String> getInstancesUuids(TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>> client) {
```
vrogach2020, line 70: Looks like a flacky check too.
I think the good scenario is:
0) set number of connections to 10
1) start 3 routers. Check that each of three clients has 10 connections via box.stat.net()
2) shut down 2 routers. Check that they are down
3) restart 2 routers. Wait. Check that only first router has 10 connections and others have 0.
4) execute establishLackingConnections()
5) check that all routers have 10 connections each.

vrogach2020 (changes requested):

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: Made it private
Static is needed for convenience

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: There will be no flacking because if we have one and the same uuid instance returning in a row, then this is a signal that there is only one router left

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: box.stat.net (). CONNECTIONS is not a very accurate indicator of connections, since connections between tarantool instances are also counted there

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: This method works correctly when using the round robin connection selection strategy. In addition, it is universal and works on any number of routers and connections.

vrogach2020 (approved):

Issue #159, opened 2021-12-02
Choose required cartridge roles to wait at connect


Totktonada:
It is possible that tarantool already listen on a TCP port, but not all necessary functions are added/initialized by a role. See https://github.com/tarantool/crud/issues/229 for example.

It is also possible that we don't grant all necessary permissions to a database user at the moment of connect (because snap/xlog is not fully applied at this moment).

I propose to add an option to wait until a set of user provided roles will be initialized and only after this consider the connection ready to use.

sharonovd:
Or simply wait for RolesConfigured cartridge state

Totktonada:
Well, that sounds even better.

PR #162, opened 2022-01-21
Refactoring for RequestRetryPolicies and classes in exceptions.errors package

wey1and:
Added refactoring for RequestRetryPolicies and classes in `exceptions.errors` package

Added new Exception called `TarantoolNoSuchProcedureException` for this message, so that the user can specify it in the retry policy
```
InnerErrorMessage:
code: 33
message: Procedure '*' is not defined
```
Closes https://github.com/tarantool/cartridge-java/issues/159

### DIFF
--- a/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
+++ b/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
@@ -31,23 +31,12 @@ public final class TarantoolRequestRetryPolicies {
     public static final long DEFAULT_ONE_HOUR_TIMEOUT = TimeUnit.HOURS.toMillis(1); //ms
 
     /**
-     * Check all known network exceptions and the user-specified exceptions
-     *
-     * @param exceptionCheck specified predicate for checking exceptions
-     * @param <T>            type of user specified predicate for checking exceptions
-     * @return predicate for checking all networks exceptions
-     */
-    public static <T extends Predicate<Throwable>> Predicate<Throwable> withRetryingNetworkErrors(T exceptionCheck) {
-        return exceptionCheck.or(TarantoolRequestRetryPolicies::isNetworkError);
-    }
-
-    /**
      * Check all known network exceptions
      *
      * @return predicate for checking all network exceptions
      */
     public static Predicate<Throwable> retryNetworkErrors() {
-        return withRetryingNetworkErrors(retryNone);
+        return TarantoolRequestRetryPolicies::isNetworkError;
     }
 
     /**
--- a/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
+++ b/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
@@ -38,7 +38,7 @@ public final class TarantoolRequestRetryPolicies {
      * @return predicate for checking all networks exceptions
      */
     public static <T extends Predicate<Throwable>> Predicate<Throwable> withRetryingNetworkErrors(T exceptionCheck) {
-        return exceptionCheck.or((TarantoolRequestRetryPolicies::isNetworkError));
+        return exceptionCheck.or(TarantoolRequestRetryPolicies::isNetworkError);
     }
 
     /**
--- a/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
+++ b/src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
@@ -4,6 +4,7 @@ import io.tarantool.driver.exceptions.TarantoolAttemptsLimitException;
 import io.tarantool.driver.exceptions.TarantoolClientException;
 import io.tarantool.driver.exceptions.TarantoolConnectionException;
 import io.tarantool.driver.exceptions.TarantoolInternalNetworkException;
+import io.tarantool.driver.exceptions.TarantoolNoSuchProcedureException;
 import io.tarantool.driver.exceptions.TarantoolTimeoutException;
 import io.tarantool.driver.utils.Assert;
 
@@ -13,7 +14,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 /**
@@ -21,41 +22,47 @@ import java.util.function.Supplier;
  *
  * @author Alexey Kuzin
  * @author Artyom Dubinin
+ * @author Oleg Kuznetsov
  */
 public final class TarantoolRequestRetryPolicies {
 
-    public static final Function<Throwable, Boolean> retryAll = t -> true;
-    public static final Function<Throwable, Boolean> retryNone = t -> false;
+    public static final Predicate<Throwable> retryAll = t -> true;
+    public static final Predicate<Throwable> retryNone = t -> false;
     public static final long DEFAULT_ONE_HOUR_TIMEOUT = TimeUnit.HOURS.toMillis(1); //ms
 
     /**
      * Check all known network exceptions and the user-specified exceptions
      *
-     * @param exceptionCheck specified callback for checking exceptions
-     * @param <T>            type of user specified callback for checking exceptions
-     * @return callback for checking all networks exceptions
+     * @param exceptionCheck specified predicate for checking exceptions
+     * @param <T>            type of user specified predicate for checking exceptions
+     * @return predicate for checking all networks exceptions
      */
-    public static <T extends Function<Throwable, Boolean>> Function<Throwable, Boolean>
-    withRetryingNetworkErrors(T exceptionCheck) {
-        return e -> {
-            boolean retryRequest = false;
-            Boolean userExceptionCheck = exceptionCheck.apply(e);
-            if (e instanceof TimeoutException ||
-                    e instanceof TarantoolConnectionException ||
-                    e instanceof TarantoolInternalNetworkException) {
-                retryRequest = true;
-            }
-            return retryRequest || userExceptionCheck;
-        };
+    public static <T extends Predicate<Throwable>> Predicate<Throwable> withRetryingNetworkErrors(T exceptionCheck) {
+        return exceptionCheck.or((TarantoolRequestRetryPolicies::isNetworkError));
     }
 
     /**
      * Check all known network exceptions
      *
-     * @return callback for checking all network exceptions
+     * @return predicate for checking all network exceptions
      */
-    public static Function<Throwable, Boolean> retryNetworkErrors() {
+    public static Predicate<Throwable> retryNetworkErrors() {
         return withRetryingNetworkErrors(retryNone);
+    }
+
+    /**
+     * Check {@link TarantoolNoSuchProcedureException}
+     *
+     * @return predicate for checking {@link TarantoolNoSuchProcedureException}
+     */
+    public static Predicate<Throwable> retryTarantoolNoSuchProcedureErrors() {
+        return throwable -> throwable instanceof TarantoolNoSuchProcedureException;
+    }
+
+    private static boolean isNetworkError(Throwable e) {
+        return e instanceof TimeoutException ||
+                e instanceof TarantoolConnectionException ||
+                e instanceof TarantoolInternalNetworkException;
     }
 
     private TarantoolRequestRetryPolicies() {
@@ -67,13 +74,12 @@ public final class TarantoolRequestRetryPolicies {
      *
      * @param <T> exception checking callback function type
      */
-    public static final class InfiniteRetryPolicy<T extends Function<Throwable, Boolean>>
-            implements RequestRetryPolicy {
+    public static final class InfiniteRetryPolicy<T extends Predicate<Throwable>> implements RequestRetryPolicy {
 
         private final long requestTimeout; //ms
         private final long operationTimeout; //ms
         private final long delay; //ms
-        private final T callback;
+        private final T predicate;
 
         /**
          * Basic constructor
@@ -81,7 +87,7 @@ public final class TarantoolRequestRetryPolicies {
          * @param requestTimeout   timeout for one retry attempt, in milliseconds
          * @param operationTimeout timeout for the whole operation, in milliseconds
          * @param delay            delay between attempts, in milliseconds
-         * @param exceptionCheck   function checking whether the given exception may be retried
+         * @param exceptionCheck   predicate checking whether the given exception may be retried
          */
         public InfiniteRetryPolicy(long requestTimeout, long operationTimeout, long delay, T exceptionCheck) {
             Assert.state(requestTimeout >= 0, "Timeout must be greater or equal than 0!");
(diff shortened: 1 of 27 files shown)


src/main/java/io/tarantool/driver/exceptions/errors/ErrorCode.java
```diff
@@ -0,0 +1,17 @@
+package io.tarantool.driver.exceptions.errors;
+
+/**
+ * Error codes used to classify errors
```
ArtDu, line 4: Please describe here where these codes come from, you can link to the code in the documentation or github

src/main/java/io/tarantool/driver/exceptions/errors/TarantoolErrors.java
```diff
@@ -138,12 +99,7 @@ private Boolean isErrorsError(String errorMessage) {
      * @see <a href="https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/error/">box_error</a>
      */
     public static class TarantoolBoxErrorFactory implements TarantoolErrorFactory {
```
ArtDu, line 101: We need to think about renaming BoxError. Because we have the logic that lua error() and tarantool box.error() generate IPROTO_NOT_OK. Those the principle of their work is the same, box.error can be caught in the same way with pcall(). And if we generate lua error() it is perceived by the connector as box.error

ArtDu: This is an incorrect assumption about lua's excellent `error()` structure. Please leave a correct link to the doc, where it says that lua `error()` is wrapped in box.error with code 32. Now a link to a specific method, but you need to the whole box.error - https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -34,19 +41,36 @@ public static void setUp() throws TimeoutException {
         PASSWORD = container.getPassword();
     }
 
+    private boolean crudProcedureIsNotDefined;
```
vrogach2020, line 44: This can be a local variable in  test_should_reconnect_ifReconnectIsInvoked()

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -34,19 +41,36 @@ public static void setUp() throws TimeoutException {
...
+            }
+        });
+        new Thread(() -> {
+            for (int i = 0; i < 10000; i++) {
+                client.space("test_space").select(Conditions.any()).join();
+            }
+        }).start();
+
+        client.eval("rawset(_G, 'crud', nil)").join();
+        Thread.sleep(500);
```
vrogach2020, line 63: Looks like flaky condition - you have retry with 5 attempts and delay 100ms.   Maybe decrease this sleep to 300 ?

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -34,19 +41,36 @@ public static void setUp() throws TimeoutException {
         PASSWORD = container.getPassword();
     }
 
+    private boolean crudProcedureIsNotDefined;
+
+    @Test
+    public void test_should_reconnect_ifReconnectIsInvoked() throws InterruptedException {
```
vrogach2020, line 47: In what line of this tescase the 'reconnect is invoked' ?

vrogach2020 (changes requested):

src/main/java/io/tarantool/driver/exceptions/errors/BoxErrorKey.java
```diff
@@ -0,0 +1,35 @@
...
+import org.msgpack.value.StringValue;
+import org.msgpack.value.ValueFactory;
+
+/**
+ * Errors keys for error message in box error
+ *
+ * @author Oleg Kuznetsov
+ */
+enum BoxErrorKey implements ErrorKey {
+    CODE("code", ValueFactory.newString("code")),
```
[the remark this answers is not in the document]
wey1and: This is necessary in order not to create a msgPack string every time, since enum values are static, which will lead to better performance.

src/main/java/io/tarantool/driver/exceptions/errors/ErrorsErrorKey.java
```diff
@@ -0,0 +1,36 @@
...
+ * Errors keys for error message from tarantool/errors library
+ *
+ * @author Oleg Kuznetsov
+ */
+enum ErrorsErrorKey implements ErrorKey {
+    LINE("line", ValueFactory.newString("line")),
+    CLASS_NAME("class_name", ValueFactory.newString("line")),
+    ERR("err", ValueFactory.newString("err")),
+    FILE("file", ValueFactory.newString("file")),
+    ERROR_MESSAGE("str", ValueFactory.newString("str")),
```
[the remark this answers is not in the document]
wey1and: These values were taken from the constants in the refactored code without changes

src/main/java/io/tarantool/driver/exceptions/errors/TarantoolErrorsParser.java
```diff
@@ -36,14 +37,12 @@ private TarantoolErrorsParser() {
...
         try {
-            Optional<TarantoolException> exception;
-            for (TarantoolErrorFactory errorFactory : errorsFactories) {
-                exception = errorFactory.create(error);
-                if (exception.isPresent()) {
-                    return exception.get();
-                }
-            }
-            return defaultFactory.create(error).get();
+            return errorsFactories.stream()
```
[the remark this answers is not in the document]
wey1and: Maybe you are right. I suggested a more compact alternative.

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: You cannot assign effectively final variables inside lambda expressions. But we can make it atomic boolean.

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
wey1and: We discussed it and talked about a few milliseconds. How many times I ran it - nothing flacked. But i will increase just to your sure.

src/main/java/io/tarantool/driver/exceptions/errors/BoxErrorKey.java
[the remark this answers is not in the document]
akudiyar: I agree with Oleg, this way it seems better.

src/main/java/io/tarantool/driver/exceptions/errors/TarantoolErrorsParser.java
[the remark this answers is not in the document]
akudiyar: Use first() + orElse() here, since you don't really need a list.

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
akudiyar: This test seems unreliable, because we don't force the conditions on the server side. It may always pass in the test env by that.

akudiyar (changes requested):
Mostly LGTM, but IT tests need more consideration.

src/main/java/io/tarantool/driver/exceptions/errors/BoxErrorKey.java
```diff
@@ -0,0 +1,35 @@
+package io.tarantool.driver.exceptions.errors;
+
+import org.msgpack.value.StringValue;
+import org.msgpack.value.ValueFactory;
+
+/**
+ * Errors keys for error message in box error
```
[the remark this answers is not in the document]
wey1and: Discussed verbally with Vladimir. We decided not to add it yet, since this was not originally, and the current changes are only refactoring.

src/main/java/io/tarantool/driver/exceptions/errors/ErrorsErrorKey.java
```diff
@@ -0,0 +1,36 @@
+package io.tarantool.driver.exceptions.errors;
+
+import org.msgpack.value.StringValue;
+import org.msgpack.value.ValueFactory;
+
+/**
+ * Errors keys for error message from tarantool/errors library
```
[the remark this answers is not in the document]
wey1and: Discussed verbally with Vladimir. We decided not to add it yet, since this was not originally, and the current changes are only refactoring.

vrogach2020 (approved):

vrogach2020:
@wey1and please do not forget to include backwards compatibility changes into changelog

ArtDu:
@wey1and Please squash commits and make one commit message with list of what has changed

src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
```diff
@@ -13,51 +14,57 @@
...
+    public static Predicate<Throwable> retryNetworkErrors() {
         return withRetryingNetworkErrors(retryNone);
     }
 
+    /**
+     * Check {@link TarantoolNoSuchProcedureException}
+     *
+     * @return predicate for checking {@link TarantoolNoSuchProcedureException}
+     */
+    public static Predicate<Throwable> retryTarantoolNoSuchProcedureErrors() {
```
akudiyar, line 58: There should be also a `withRetryingNoSuchProcedureErrors` compartment for this function, otherwise the user will not be able to extend this built-in checker with custom error checkers.

src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
```diff
@@ -150,7 +150,7 @@ private TarantoolConnectionManager connectionManager() {
 
     @Override
     public boolean establishLackingConnections() {
-        return connectionManager.establishLackingConnections();
+        return connectionManager().establishLackingConnections();
```
akudiyar, line 153: Please rename the function in a separate PR before making a release. We shouldn't show any implementation details in API (corresponds to the Open-Closed principle in SOLID).

src/main/java/io/tarantool/driver/exceptions/TarantoolNoSuchProcedureException.java
```diff
@@ -0,0 +1,19 @@
+package io.tarantool.driver.exceptions;
+
+/**
+ * An exception occurs when the procedure has not (yet) been defined.
```
akudiyar, line 4:
```suggestion
 * Corresponds to an exception that occurs when the procedure has not (yet) been defined in the Tarantool instance.
```

src/main/java/io/tarantool/driver/exceptions/TarantoolNoSuchProcedureException.java
```diff
@@ -0,0 +1,19 @@
...
+ * <p></p>
+ * For example: this exception can be raised when a role is not up yet or being reloaded at the moment.
```
akudiyar, line 6:
```suggestion
 * For example: this exception can be raised when a Cartridge role is not up yet or is being reloaded at the moment.
```

src/main/java/io/tarantool/driver/exceptions/TarantoolNoSuchProcedureException.java
```diff
@@ -0,0 +1,19 @@
...
+ * This can happen when the tarantool 'router' node is restarted or hot-reloaded but the cluster is under
+ * load at the moment. If CRUD module is used in the cluster then 'cartridge.roles.crud-router' may
+ * have not been initialized yet and all requests to CRUD will fail with this exception.
+ *
+ * You can use this exception as a sign that it's worth trying a few more times until the role is up.
```
akudiyar, line 11:
```suggestion
 * You can use this exception when you are sure that it's worth trying a few more times until the role is up.
```

src/main/java/io/tarantool/driver/exceptions/errors/TarantoolErrors.java
```diff
@@ -123,27 +83,22 @@ private Boolean isNetworkError(String err) {
...
+     * @see <a href="https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_error/">box_error</a>
      */
     public static class TarantoolBoxErrorFactory implements TarantoolErrorFactory {
-        private static final StringValue CODE = ValueFactory.newString("code");
-        private static final StringValue BASE_TYPE = ValueFactory.newString("base_type");
-        private static final StringValue TYPE = ValueFactory.newString("type");
-        private static final StringValue MESSAGE = ValueFactory.newString("message");
-        private static final StringValue TRACE = ValueFactory.newString("trace");
-        private static final Long ERROR_INDICATOR_OFFSET = 32768L;
+        private static final int ERROR_INDICATOR_OFFSET = 32768;
```
akudiyar, line 101: Is omitting "L" here safe? Did you test it?

akudiyar (changes requested):

src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
wey1and: Yes, thank you, we remember this.

src/main/java/io/tarantool/driver/api/retry/TarantoolRequestRetryPolicies.java
wey1and: Discussed with Alexey in correspondence. We decided that the comment is not relevant, since the Predicate interface provides the necessary API.

ArtDu:
nit: 

The PR title does not reveal what is in the PR. It seems better not to use the word refactoring, because there was not only a refactoring, but also changes in API of the `exceptions.errors` package. And since the external behavior of the code will change, then you can not use this word, it seems better to use a `CHANGE` or something like that. 

Just this will be a generalized version of what is happening in PR. And in the description, you can just write that a refactoring has taken place, the API has changed and a new exception has been added and why it was added. And I expect to see this also in the final squashed commit message.

ArtDu (dismissed):
LGTM

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
```diff
@@ -123,7 +127,12 @@ public static void initClient() {
 
         ClusterTarantoolTupleClient clusterClient = new ClusterTarantoolTupleClient(
                 config, getClusterAddressProvider());
-        client = new ProxyTarantoolTupleClient(clusterClient);
+
+        client = new RetryingTarantoolTupleClient(new ProxyTarantoolTupleClient(clusterClient),
```
akudiyar, line 131: nit: Why do we still use an old approach here and not the new shiny Builder?

src/test/java/io/tarantool/driver/integration/ReconnectIT.java
```diff
@@ -34,19 +43,40 @@ public static void setUp() throws TimeoutException {
...
+        new Thread(() -> {
+            for (int i = 0; i < 10000; i++) {
+                client.space("test_space").select(Conditions.any()).join();
+            }
+        }).start();
+
+        client.eval("rawset(_G, 'crud', nil)").join();
+        Thread.sleep(500);
+        client.eval("rawset(_G, 'crud', require('crud'))").join();
+        Thread.sleep(500);
```
akudiyar, line 71: Do not get why we need the second sleep. I'd rather write this test in another way: 
1) Set the crud to nil and loop with small sleeps until a target error occurs
2) Set the crud back and loop with small sleeps until the call passes without errors or fail the test after a timeout.

akudiyar (changes requested):

vrogach2020 (approved):

akudiyar (approved):
LGTM.

BTW there is a `truncate_space()` test failing on 1.10 -- we need to fix it.

ArtDu (approved):

vrogach2020:
@akudiyar  Created an issue on truncate
#164

Issue #163, opened 2022-01-31
Initiate reconnect for DiscoveryClusterAddressProvider


wey1and:
When receiving a status that indicates to us that the node is unavailable from `ServerNodeInfo` in `DiscoveryClusterAddressProvider`, the driver does not initiate a reconnect, thereby not taking the current list of addresses. It is necessary to initiate reconnection to the nodes in case of failure of at least one node.

PR #167, opened 2022-02-15
Added refresh callback to address provider

wey1and:
Closes https://github.com/tarantool/cartridge-java/issues/163

### DIFF
--- a/src/main/java/io/tarantool/driver/api/TarantoolClient.java
+++ b/src/main/java/io/tarantool/driver/api/TarantoolClient.java
@@ -78,5 +78,5 @@ public interface TarantoolClient<T extends Packable, R extends Collection<T>>
      *
      * @return returns true if the establishing process has been started, else false
      */
-    boolean establishLackingConnections();
+    boolean refresh();
 }
--- a/src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java
+++ b/src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java
@@ -10,10 +10,21 @@ import java.util.Collection;
  */
 public interface TarantoolClusterAddressProvider extends AutoCloseable {
     /**
-     * The the collection of Tarantool server nodes which belong to the same cluster
+     * The collection of Tarantool server nodes which belong to the same cluster
+     *
      * @return collection of {@link TarantoolServerAddress}
      */
     Collection<TarantoolServerAddress> getAddresses();
+
+    /**
+     * Specify callback for refreshing connections to addresses.
+     * <p>
+     * For example: you can run it when you want saying about your Tarantool server addresses is changed
+     *
+     * @param runnable callback for running refresh connections
+     */
+    default void setRefreshCallback(Runnable runnable) {
+    }
 
     @Override
     default void close() {
--- a/src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java
+++ b/src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java
@@ -19,7 +19,7 @@ public interface TarantoolClusterAddressProvider extends AutoCloseable {
     /**
      * Specify callback for refreshing connections to addresses.
      * <p>
-     * For example: you can run it when you want saying about your Tarantool server addresses is changed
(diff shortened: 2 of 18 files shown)


vrogach2020:
Please add a description and/or issue link

src/main/java/io/tarantool/driver/core/connection/events/EventManager.java
```diff
@@ -0,0 +1,37 @@
+package io.tarantool.driver.core.connection.events;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class EventManager<E extends Enum<E>> {
```
vrogach2020, line 8: I suggest to replace generic by event interface. I don't see any reasons to parameterize event type. Enum can implement the event interface so you can keep event list in enum

src/main/java/io/tarantool/driver/api/TarantoolClient.java
```diff
@@ -73,10 +75,12 @@
      */
     TarantoolConnectionListeners getConnectionListeners();
 
+    EventManager<ConnectionManagerEvent> getEventManager();
```
vrogach2020, line 78: There are no cases yet that require this API to be available to customers. I think we should keep internal and create eventManager in discoveryAddressProvider where it solves the problem.

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -42,42 +45,32 @@
...
      * @param config              Tarantool client config
      * @param connectionFactory   connection factory
      * @param connectionListeners connection listeners
+     * @param eventManager        event manager
      */
     public AbstractTarantoolConnectionManager(TarantoolClientConfig config,
                                               TarantoolConnectionFactory connectionFactory,
-                                              TarantoolConnectionListeners connectionListeners) {
+                                              TarantoolConnectionListeners connectionListeners,
+                                              EventManager<ConnectionManagerEvent> eventManager) {
```
vrogach2020, line 64: As discussed - we are going to refactor and split this class. So I suggest not to add new logic here and inject eventmanager into TarantoolClusterConnectionManager.

vrogach2020 (changes requested):

src/main/java/io/tarantool/driver/core/connection/EventListener.java
```diff
@@ -0,0 +1,11 @@
+package io.tarantool.driver.core.connection;
+
+/**
+ * Basic event listener interface
+ *
+ * @author Oleg Kuznetsov
+ */
+public interface EventListener {
+
+    void update();
```
vrogach2020, line 10: Seems useful to add event parameter
```update(EventType event)```

**Review by vrogach2020 — CHANGES_REQUESTED** [review]

File: src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java [hunk 5]
```diff
@@ -10,11 +13,22 @@
...
      */
     Collection<TarantoolServerAddress> getAddresses();
 
+    /**
+     * Returns event manager for notifying connection manager about external events
+     * e.g. server address changed
+     *
+     * @return event manager
+     */
+    default EventManager getEventManager() {
```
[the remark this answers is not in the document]
vrogach2020: UPD Discussed - replace with setCallback

src/main/java/io/tarantool/driver/api/TarantoolClusterAddressProvider.java
```diff
@@ -10,11 +10,22 @@
...
+     * The collection of Tarantool server nodes which belong to the same cluster
+     *
      * @return collection of {@link TarantoolServerAddress}
      */
     Collection<TarantoolServerAddress> getAddresses();
 
+    /**
+     * Specify callback for refreshing connections to addresses.
+     * <p>
+     * For example: you can run it when you want saying about your Tarantool server addresses is changed
```
vrogach2020, line 22:
```suggestion
     * For example: can be executed when provider detects changes in the list of server addresses 
```

Issue #186, opened 2022-03-21
TarantoolNullField.hashCode causes stackoverflow


vrogach2020:
Method implementation
```
  @Override
    public int hashCode() {
        return Objects.hashCode(this);
    }
```

Objects.hashCode implementation
```
    public static int hashCode(Object o) {
        return o != null ? o.hashCode() : 0;
    }
```
This obviously causes infinite recursion for non-null objects

PR #190, opened 2022-03-29
 Fix recursion in TarantoolNullField.hashCode method

dkasimovskiy:
Use super.hashCode instead of Objects.hashCode
Fix formatting in pom.xml
Closes https://github.com/tarantool/cartridge-java/issues/186

### DIFF
--- a/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
+++ b/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
@@ -44,6 +44,6 @@ public final class TarantoolNullField implements TarantoolField {
 
     @Override
     public int hashCode() {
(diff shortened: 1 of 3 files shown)


src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
```diff
@@ -44,6 +44,6 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(this);
+        return Objects.hashCode(this.toString());
```
wey1and, line 47: This won't fix the stack overflow since the `toString()` method also calls `hashСode()`

wey1and (changes requested):

wey1and (approved):

PR #192, opened 2022-03-30
Prevent multiple call of TarantoolClusterDiscoveryConfig.Builder().withEndpoint()

### LINKED ISSUE #173: Discovery endpoint is overwritten when withEndpoint() called multiple times
See TarantoolClusterDiscoveryConfig.Builder#withEndpoint.
If this method is called twice it will overwrite previous endpoint setting. 
That leads to incorrect configuration that is not checked in runtime.

Now the it's impossible to specify more that 1 endpoint. 
We need to throw an exception or to support multiple endpoints in config.

dkasimovskiy:
Throw exception if TarantoolClusterDiscoveryConfig.Builder().withEndpoint() called multiple times
Closes #173

### DIFF
--- a/src/test/java/io/tarantool/driver/cluster/TarantoolClusterDiscoveryConfigTest.java
+++ b/src/test/java/io/tarantool/driver/cluster/TarantoolClusterDiscoveryConfigTest.java
@@ -1,0 +1,39 @@
+package io.tarantool.driver.cluster;
+
+import io.tarantool.driver.exceptions.TarantoolClientException;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Dmitry Kasimovskiy
+ */
+public class TarantoolClusterDiscoveryConfigTest {
+
+    @Test
+    public void test_withEndpoint_shouldReturnConfig() {
+        // given
+        TarantoolClusterDiscoveryEndpoint endpoint = new TarantoolClusterDiscoveryEndpoint() { };
+
+        // when
+        TarantoolClusterDiscoveryConfig cfg = new TarantoolClusterDiscoveryConfig.Builder()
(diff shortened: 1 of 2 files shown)


src/test/java/io/tarantool/driver/cluster/TarantoolClusterDiscoveryConfigTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * @author Dmitry Kasimovskiy
+ */
+public class TarantoolClusterDiscoveryConfigTest {
+
+    private static class FakeTarantoolClusterDiscoveryEndpoint implements TarantoolClusterDiscoveryEndpoint {
```
wey1and, line 13: You can use empty class `new TarantoolClusterDiscoveryEndpoint(){}`, `BinaryClusterDiscoveryEndpoint` or `HTTPClusterDiscoveryEndpoint` instead.

wey1and: You can use empty class `new TarantoolClusterDiscoveryEndpoint(){}`, `BinaryClusterDiscoveryEndpoint` or `HTTPClusterDiscoveryEndpoint` instead.

wey1and (changes requested):

src/test/java/io/tarantool/driver/cluster/TarantoolClusterDiscoveryConfigTest.java
dkasimovskiy: Done

wey1and (approved):

wey1and (approved):

vrogach2020 (approved):

PR #199, opened 2022-04-08
Add support for IPv6

wey1and:


src/main/java/io/tarantool/driver/api/TarantoolServerAddress.java
```diff
@@ -3,20 +3,19 @@
...
- *
+ * For using IPv6 address, use this constructor {@link TarantoolServerAddress#TarantoolServerAddress(InetSocketAddress)}
  * @author Sergey Volgin
  */
 public class TarantoolServerAddress implements Serializable {
     private static final long serialVersionUID = 7327851568010264254L;
     private static final String DEFAULT_HOST = "127.0.0.1";
     private static final int DEFAULT_PORT = 3301;
 
+    private final InetSocketAddress inetSocketAddress;
```
Totktonada, line 18: Why don't just pass IPv6 address in the host field / `TarantoolServerAddress()` parameter in square brackets?

wey1and: Because it's not working in current implementation of main constructor, and it will throw IllegalArgumentException.
It's can to be implemented, but we need to rewrite this constructor.
This PR is first step for supporting IPv6.

Totktonada: Okay, got it.

wey1and: I added this. Now we can pass IPv6 address in any constructor.

src/main/java/io/tarantool/driver/api/TarantoolServerAddress.java
```diff
@@ -37,44 +39,98 @@ public TarantoolServerAddress(final String host) {
...
+            }
 
-        int portNumber = port;
-        if (addressParts.length == 2) {
+            int portIdx = host.indexOf("]:");
+            if (portIdx != -1) {
+                if (port != DEFAULT_PORT) {
+                    throw new IllegalArgumentException("Can't specify port in construct and via host");
+                }
+                portToUse = Integer.parseInt(host.substring(portIdx + 2));
```
vrogach2020, line 83: Please create a new ctor TarantoolServerAddress(String) for the case when host and port is specified in a single string.

vrogach2020 (changes requested):

src/main/java/io/tarantool/driver/api/TarantoolServerAddress.java
```diff
@@ -6,19 +6,21 @@
...
 
 /**
  * Represents the location of a Tarantool server - server name and port number
  *
  * @author Sergey Volgin
+ * @author Oleg Kuznetsov
  */
 public class TarantoolServerAddress implements Serializable {
     private static final long serialVersionUID = 7327851568010264254L;
     private static final String DEFAULT_HOST = "127.0.0.1";
```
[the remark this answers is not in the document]
vrogach2020: Discussed: move to default ctor

src/main/java/io/tarantool/driver/api/TarantoolServerAddress.java
```diff
@@ -3,78 +3,113 @@
...
+            }
 
-        if (addressParts.length > 2) {
-            throw new IllegalArgumentException(String.format("Invalid host name: %s", host));
+            int portIdx = host.indexOf("]:");
+            if (portIdx != -1) {
+                try {
+                    portToUse = Integer.parseInt(host.substring(portIdx + 2));
+                } catch (NumberFormatException e) {
+                    throw new IllegalArgumentException(String.format("Invalid host name: %s", host));
```
vrogach2020, line 56: Error is thrown because of port but no host. 
Maybe rename host to address in error text or complain on port but not hostname

vrogach2020 (approved):

PR #201, opened 2022-04-08
PoC: Cache objects in TarantoolField

### LINKED ISSUE #198: RFC: Cache search result of `mapper.getValueConverter`, `mapper.fromValue` in TarantoolFieldImpl
### Introduction

TarantoolField only stores raw MSGPACK values (**value** field) and each time we call the heavy converter lookup operation(`mapper.fromValue`, `mapper.getValueConverter`) for the stored **value** for the operation `getValue`, `canConvertValue`

https://github.com/tarantool/cartridge-java/blob/54ea4be461204b103f2dd41e9e2dc960011b27cc/src/main/java/io/tarantool/driver/core/tuple/TarantoolFieldImpl.java#L16-L84

### mapper.getValueConverter
We cannot store the result of this operation without qualifying the **value** class, this is because this method has a generic type _entityClass_ and returns that generic type in returnType _Optional<ValueConverter<**V**, O>>_:

https://github.com/tarantool/cartridge-java/blob/54ea4be461204b103f2dd41e9e2dc960011b27cc/src/main/java/io/tarantool/driver/mappers/MessagePackValueMapper.java#L48-L56

And since TarantoolField does not have a generic type for entityClass, we cannot specify the value for the function call value without adding generic in class signature.

Without qualification, we get this type `Optional<? extendsValueConverter<? extends Value, ?>>`, but we cannot use the fromValue method in this valueConverter. it expects this type `? extends Value` and we can only pass null (The Get and Put Principle) see **[this](https://stackoverflow.com/questions/2723397/what-is-pecs-producer-extends-consumer-super)** for details.

The only thing we can use is to add generic to the method, then we can use this converter within this method. Therefore I suggest not to cache `valueConverter` as we cannot, without adding a generic for the entire class, but cache the result of calling `valueConverter.fromValue` like this:

```java
 public <V extends Value> boolean
    canConvertValueAndCacheResult(Class<V> entityClass, Class<?> targetClass, MessagePackValueMapper mapper) {
        if (convertedValue.isPresent()) {
            return true;
        }
        Optional<? extends ValueConverter<V, ?>> converter = mapper.getValueConverter(entityClass, targetClass);
        if (converter.isPresent()) {
            convertedValue = Optional.of(converter.get().fromValue((V) value));
            return true;
        }
        return false;
    }
```
We can do this because **value** is only initialized in constructors.

### mapper.fromValue 

To do this, I suggest simply saving the result of the operation

```java
@Override
    @SuppressWarnings("unchecked")
    public <O> O getValue(Class<O> targetClass, MessagePackValueMapper mapper) {
        if (value == null) {
            return null;
        }
        if (convertedValue.isPresent()) {
            return (O) convertedValue.get();
        }
        if (value instanceof Value) {
            convertedValue = Optional.of(mapper.fromValue((Value) value, targetClass));
            return (O) convertedValue.get();
        }
        if (value.getClass().isAssignableFrom(targetClass)) {
            return (O) value;
        }
        throw new UnsupportedOperationException(
                String.format("Cannot convert field value of type %s to type %s", value.getClass(), targetClass));
    }

    @Override
    public Object getValue(MessagePackValueMapper mapper) {
        if (convertedValue.isPresent()) {
            return convertedValue.get();
        }
        if (value instanceof Value) {
            convertedValue = Optional.of(mapper.fromValue((Value) value));
            return convertedValue;
        }
        return value;
    }
```

ArtDu:
Closes #198

### DIFF
--- a/src/main/java/io/tarantool/driver/core/tuple/TarantoolFieldImpl.java
+++ b/src/main/java/io/tarantool/driver/core/tuple/TarantoolFieldImpl.java
@@ -3,19 +3,27 @@ package io.tarantool.driver.core.tuple;
 import io.tarantool.driver.api.tuple.TarantoolField;
 import io.tarantool.driver.mappers.MessagePackObjectMapper;
 import io.tarantool.driver.mappers.MessagePackValueMapper;
+import io.tarantool.driver.mappers.ValueConverter;
 import org.msgpack.value.Value;
 import org.msgpack.value.ValueFactory;
 
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Basic tuple field implementation
  *
  * @author Alexey Kuzin
+ * @author Artyom Dubinin
  */
 public class TarantoolFieldImpl implements TarantoolField {
 
-    private Object value;
+    private final Object value;
+
+    private Optional<Object> convertedValueWithoutTargetType = Optional.empty();
+
+    private Optional<Class<?>> lastConvertedTargetClass = Optional.empty();
+    private Object lastConvertedValue;
 
     /**
      * Deserializing constructor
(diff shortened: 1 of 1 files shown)


src/main/java/io/tarantool/driver/core/tuple/TarantoolFieldImpl.java
```diff
@@ -56,8 +64,14 @@ public <O> O getValue(Class<O> targetClass, MessagePackValueMapper mapper) {
         if (value == null) {
             return null;
         }
+        if (lastConvertedTargetClass.isPresent() && lastConvertedTargetClass.get().equals(targetClass)) {
```
akudiyar, line 67: This will break inevitably because we use the same target container classes for different calls and even inside the same call. This was my mistake when I implemented caching for the first time.

akudiyar (changes requested):

Issue #180, opened 2022-03-10
Add TLS/SSL support


no1seman:
Since TEE SDK r455 (master branch) released with base support of IPROTO over TLS/SSL need to add TLS/SSL support to Java-connector.

PR #202, opened 2022-04-13
Added support SSL/TLS

wey1and:
Closes https://github.com/tarantool/cartridge-java/issues/180


src/test/java/io/tarantool/driver/integration/SslClientIT.java
```diff
@@ -0,0 +1,44 @@
...
+import io.tarantool.driver.exceptions.TarantoolClientException;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import javax.net.ssl.SSLException;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@Testcontainers
```
vrogach2020, line 19: Это не нужно, у нас тестконтейнер статический шареный - им не надо управлять извне

src/main/java/io/tarantool/driver/exceptions/TarantoolSSLException.java
```diff
@@ -0,0 +1,7 @@
+package io.tarantool.driver.exceptions;
+
+public class TarantoolSSLException extends TarantoolException {
```
ArtDu, line 3: Where is this excepion used? I haven't found any mentions

wey1and: This is for the future

ArtDu:
Maybe can you make tests based on `jmh`, as I did for mine PR? 
https://github.com/tarantool/cartridge-java/pull/204

We will find out how much slower  ssl is compared to regular iproto

akudiyar:
Please add a brief description of the changes -- what was changed, why this change is needed.

src/main/java/io/tarantool/driver/api/TarantoolClientConfig.java
```diff
@@ -284,7 +304,19 @@ public Builder withMessagePackMapper(MessagePackMapper mapper) {
...
+        }
+
+        /**
+         * Specify SslContext with settings for establishing SSL/TLS connection between Tarantool
+         *
+         * @param sslContext {@link SslContext} instance
+         * @return builder
+         */
+        public Builder withSslContext(SslContext sslContext) {
+            Assert.notNull(sslContext, "Specified in client config SslContext must not be null");
```
akudiyar, line 318: Nit: "SslContext must not be null" is enough.

src/main/java/io/tarantool/driver/exceptions/TarantoolSSLException.java
akudiyar: Please don't add any not used code.

src/test/java/io/tarantool/driver/integration/SslClientIT.java
```diff
@@ -0,0 +1,97 @@
...
+        final TarantoolServerAddress address =
+                new TarantoolServerAddress(container.getHost(), container.getMappedPort(3301));
+        final SimpleTarantoolCredentials credentials =
+                new SimpleTarantoolCredentials(container.getUsername(), container.getPassword());
+
+        final ClusterTarantoolTupleClient clientWithSsl = makeClientWithSsl(address, credentials);
+        final ClusterTarantoolTupleClient clientWithoutSsl = makeClient(address, credentials);
+
+        assertDoesNotThrow(clientWithoutSsl::getVersion);
+        assertThrows(TarantoolClientException.class, clientWithSsl::getVersion);
```
akudiyar, line 56: It would be nice to check the exception message here as well.

src/test/resources/org/testcontainers/containers/ssl/ssl_server.lua
```diff
@@ -0,0 +1,11 @@
+box.cfg { listen = {
```
akudiyar, line 1: This server script doesn't seem to be used anywhere. Isn't there a mistake in the tests?

src/main/java/io/tarantool/driver/api/TarantoolClientConfig.java
```diff
@@ -284,7 +304,19 @@ public Builder withMessagePackMapper(MessagePackMapper mapper) {
...
+            return this;
+        }
+
+        /**
+         * Specify SslContext with settings for establishing SSL/TLS connection between Tarantool
+         *
+         * @param sslContext {@link SslContext} instance
+         * @return builder
+         */
+        public Builder withSslContext(SslContext sslContext) {
```
akudiyar, line 317: I would suggest the following:
Add a method that adds a "secure" flag like this: https://github.com/beyondglow/spring-sourcecode/blob/8c983b1d818739ed62677cabbda79c827339afe7/spring-web/src/main/java/org/springframework/http/client/Netty4ClientHttpRequestFactory.java#L198. This will allow users to have an SSL config with disabling the SSL connection on demand in some code places without ugly setting the context to `null`.
Also, it may be useful to add an ability to set up a default SSL context if this `secure` flag is set but the context is not set up. This will ease the life of users and they won't have to learn the Netty docs for just testing the SSL connection.

akudiyar (changes requested):
Please add examples of using the new options and methods in the README and javadocs.

src/main/java/io/tarantool/driver/exceptions/TarantoolSSLException.java
ArtDu: I thought that "future" exception will be used in this PR. @wey1and correct me if i'm wrong

src/main/java/io/tarantool/driver/api/TarantoolClientConfig.java
wey1and: Thanks for your suggestion! We will add this in next PR's.

vrogach2020 (approved):

vrogach2020:
@wey1and please create issues on follow-ups from @akudiyar and @ArtDu

src/main/java/io/tarantool/driver/api/TarantoolClientConfig.java
wey1and: https://github.com/tarantool/cartridge-java/issues/220

PR #204, opened 2022-04-14
Change value converters searching

### LINKED ISSUE #178: RFC: ValueConverters search optimization
### What algorithm is now:

There is a search in all superclasses and interfaces for the presence of a converter for the incoming MessagePack object.
* Where a MessagePack of a certain type appears in general? - `MessagePack.newDefaultUnpacker(in)` is decoded (see `MessagePackFrameDecoder`) -> `unpacker.unpackValue()`:

```java
public ImmutableValue unpackValue()
        throws IOException
{
    MessageFormat mf = getNextFormat();
    switch (mf.getValueType()) {
        ...
        case FLOAT:
            return ValueFactory.newFloat(unpackDouble());
        case STRING: {
            int length = unpackRawStringHeader();
            return ValueFactory.newString(readPayload(length), true);
        }
...
public static ImmutableStringValue newString(byte[] b, boolean omitCopy)
{
    if (omitCopy) {
        return new ImmutableStringValueImpl(b);
    }
    else {
        return new ImmutableStringValueImpl(Arrays.copyOf(b, b.length));
    }
}
```
Those fixed values are always returned

* What converters are registered for the incoming object? - the basic interface of the object that can come to us

```java
defaultSimpleTypesMapper = new DefaultMessagePackMapper.Builder()
            // converters for primitive values
            .withConverter(StringValue.class, Character.class, new DefaultCharacterConverter())
            .withConverter(StringValue.class, String.class, new DefaultStringConverter())
            .withValueConverter(IntegerValue.class, Float.class, new DefaultIntegerValueToFloatConverter())
            .withValueConverter(IntegerValue.class, Double.class, new DefaultIntegerValueToDoubleConverter())
            .withConverter(IntegerValue.class, Long.class, new DefaultLongConverter())
            .withConverter(IntegerValue.class, Integer.class, new DefaultIntegerConverter())
            //TODO: add this when will it be resolved https://github.com/tarantool/cartridge-java/issues/118
            // .withConverter(IntegerValue.class, Short.class, new DefaultShortConverter())
            .withConverter(BinaryValue.class, byte[].class, new DefaultByteArrayConverter())
            .withConverter(BooleanValue.class, Boolean.class, new DefaultBooleanConverter())
            .withValueConverter(FloatValue.class, Short.class, new DefaultFloatValueToShortConverter())
            .withValueConverter(FloatValue.class, Long.class, new DefaultFloatValueToLongConverter())
            .withValueConverter(FloatValue.class, Integer.class, new DefaultFloatValueToIntegerConverter())
            .withConverter(FloatValue.class, Float.class, new DefaultFloatConverter())
            .withConverter(FloatValue.class, Double.class, new DefaultDoubleConverter())
            .withConverter(ExtensionValue.class, UUID.class, new DefaultUUIDConverter())
            .withConverter(ExtensionValue.class, BigDecimal.class, new DefaultBigDecimalConverter())
            .withValueConverter(new DefaultNilConverter())
            .build();
```

* How is the search for a converter? A recursive search is done for each class, we go through all its superclasses and all interfaces

```java
private <T> Optional<T> findConverter(Class<?> objectClass, Function<String, Optional<T>> getter) {
    Optional<T> converter = getter.apply(objectClass.getTypeName());
    if (!converter.isPresent() && objectClass.getSuperclass() != null) {
        converter = findConverter(objectClass.getSuperclass(), getter);
    }
    if (!converter.isPresent()) {
        for (Class<?> iface : objectClass.getInterfaces()) {
            converter = findConverter(iface, getter);
            if (converter.isPresent()) {
                break;
            }
        }
    }
    return converter;
}
```
    
For example, `ImmutableStringValueImpl` came:
  1. Search for a converter for `ImmutableStringValueImpl` - not found
  2. Search for a converter for the superclass `AbstractImmutableRawValue` - not found
  3. Search for a converter for the superclass `AbstractImmutableValue` - not found
  4. Search for a converter for the `ImmutableValue` interface - not found
  5. Search for a converter for the `Value` interface - not found
  6. Search for a converter for the `ImmutableRawValue` interface - not found
  7. Search for converter for `RawValue` interface - not found
  8. Search for a converter for the `Value` interface - not found
  9. Search for a converter for the `ImmutableValue` interface - not found
  10. Search for a converter for the `ImmutableStringValue` interface - not found
  11. Search a converter for the `StringValue` interface - found

Those for each incoming line, we do such a search.

### First suggestion for optimization

An optimization suggestion is to add caching values to avoid doing this lookup every time:

For example add this converter:
```java
.withConverter(ImmutableStringValueImpl.class, String.class, new DefaultImmutableStringImplConverter())
```
```java
public class DefaultImmutableStringImplConverter implements
        ValueConverter<ImmutableStringValueImpl, String>, ObjectConverter<String, ImmutableStringValueImpl> {

    private static final long serialVersionUID = 20200708L;

    @Override
    public ImmutableStringValueImpl toValue(String object) {
        return new ImmutableStringValueImpl(object);
    }

    @Override
    public String fromValue(ImmutableStringValueImpl value) {
        return value.asString();
    }
}
```

ArtDu:
* Added perfomance test based on JMH
* Refactored converters magic for this purpose (point one) https://github.com/tarantool/cartridge-java/issues/178#issuecomment-1102812542 .  In short, the converters containing the implementing and ValueConverter and ObjectConveter are divided into separate classes for ease of use and in order to be able to register a `valueConverter` of this type `ImmutableStringValueImpl -> String`, and an `objectConverter` of this type `String` -> `StringValue`. More details in the code 😁
*  Also, in order for the names to be consistent, the names of the converter classes were changed
*  And additional package directories are added, because we had more than 50 files in the mappers directory and not all mappers, some converters
* Removed reflection searching to interface of MsgPack Lib enitities (point two) https://github.com/tarantool/cartridge-java/issues/178#issuecomment-1102812542  

Closes #178


vrogach2020:
It would be nice to see some perf tests results here on simple and complex types

akudiyar:
@ArtDu please add a brief description of the changes -- what was changed, why this change is needed.

pom.xml
```diff
@@ -213,6 +213,16 @@
             <version>5.7.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.35</version>
```
akudiyar, line 219: This needs to be in the test scope

pom.xml
```diff
@@ -213,6 +213,16 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.35</version>
```
akudiyar, line 224: Test scope as above

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
```diff
@@ -90,7 +90,7 @@ <V extends Value, O, T extends ValueConverter<V, O> & ObjectConverter<O, V>> Mes
      * @see io.tarantool.driver.mappers.DefaultMessagePackMapper#registerValueConverter(Class, Class, ValueConverter)
      */
     <V extends Value, O> MessagePackMapperBuilder withValueConverter(Class<V> valueClass, Class<O> objectClass,
-                                                                     ValueConverter<V, O> converter);
+                                                                     ValueConverter<?, O> converter);
```
akudiyar, line 93: This is a bad change actually. Why do we need it? I think we need a more precise generic definition, like `? extends V`, for example.

src/main/java/io/tarantool/driver/mappers/AbstractResultMapper.java
```diff
@@ -42,7 +43,7 @@ public <V extends Value, O> O fromValue(V v, Class<O> targetClass) throws Messag
     @Override
     public <V extends Value, O> void registerValueConverter(Class<V> valueClass,
                                                             Class<? extends O> objectClass,
-                                                            ValueConverter<V, ? extends O> converter) {
+                                                            ValueConverter<?, ? extends O> converter) {
```
akudiyar, line 46: Same as above, wrong generic definition

src/main/java/io/tarantool/driver/mappers/DefaultExtensionValueToUUIDConverter.java
```diff
@@ -0,0 +1,59 @@
+package io.tarantool.driver.mappers;
```
akudiyar, line 1: Don't get why this file appears as new. Would you mind describing these changes?

src/main/java/io/tarantool/driver/mappers/DefaultBigIntegerValueToFloatConverter.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Default {@link IntegerValue} to {@link Float} converter
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+public class DefaultBigIntegerValueToFloatConverter implements ValueConverter<IntegerValue, Float> {
+
+    private static final long serialVersionUID = 5249538902877730493L;
```
akudiyar, line 13: Is this an original version ID? We used to put dates as serial version IDs.

akudiyar (changes requested):

src/main/java/io/tarantool/driver/mappers/AbstractResultMapper.java
ArtDu: Changed to
```java
public <V extends Value, O> void registerValueConverter(Class<V> valueClass,
                                                      Class<? extends O> objectClass,
                                                      ValueConverter<V, ? extends O> converter) {


```

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
ArtDu: Changed to
```java
<V extends Value, O> MessagePackMapperBuilder withValueConverter(Class<? extends V> valueClass, Class<O> objectClass,
                                                                     ValueConverter<V, O> converter);


```

src/main/java/io/tarantool/driver/mappers/DefaultExtensionValueToUUIDConverter.java
ArtDu: In short, the converters containing the implementing and ValueConverter and ObjectConveter are divided into separate classes for ease of use and in order to be able to register a `valueConverter` of this type `ImmutableStringValueImpl -> String`, and an `objectConverter` of this type `String -> StringValue`

src/main/java/io/tarantool/driver/mappers/DefaultBigIntegerValueToFloatConverter.java
ArtDu: Fixed

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
```diff
@@ -1,8 +1,8 @@
 package io.tarantool.driver.api;
 
 import io.tarantool.driver.mappers.MessagePackMapper;
-import io.tarantool.driver.mappers.ObjectConverter;
-import io.tarantool.driver.mappers.ValueConverter;
+import io.tarantool.driver.mappers.converters.ObjectConverter;
+import io.tarantool.driver.mappers.converters.ValueConverter;
```
ArtDu, line 5: I don't mind leaving converters in the same folders to avoid breakin changes, but it seems better to separate converters and mappers, given that there are more than 50 files in one package

vrogach2020 (approved):

ArtDu:
Should test perf of converting with TarantoolTuple and getting with targetType

```java
TarantoolTuple tuple = ...
tuple.getObject("fieldName", Integer.class)
```

akudiyar (approved):
LGTM.
However I **highly** recommend to do at least local testing of the new changes with SpringData before merging. 

P.S. The new converters should go in a follow-up PR. The GH review interface is already completely broken at this number of changes +_+

ArtDu:
> LGTM. However I **highly** recommend to do at least local testing of the new changes with SpringData before merging.
> 
> P.S. The new converters should go in a follow-up PR. The GH review interface is already completely broken at this number of changes +_+

Yeah, I tested snapshot with cartridge-springdata, we have to rewrite registration of `TarantoolAutoConverter` but everything else looks fine

PR #206, opened 2022-04-15
Added review checklist and PR template

vrogach2020:


wey1and (approved):

ArtDu (dismissed):

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+What has been done? Why? What problem is being solved?
+
+I didn't forget about
+
+- [ ] Tests
+- [ ] Changelog
+- [ ] Documentation
```
akudiyar, line 7: I think that we are lacking doc update instructions (javadocs are pushed to a special branch that contains only the javadocs).

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
+# Code review checklist for java projects
```
akudiyar, line 1: I think that we need some limits for the PR size and some recommendations for test coverage (what can be covered with unit tests, what needs to be covered with integration tests), as well as recommendations for the Javadocs, since they are very important for the driver usage.

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
...
+- Added changes to CHANGELOG
+- No changes in files not related to the task
+  (except for a simple refactoring that obviously brings benefits and does not complicate the review)
+- No obvious flaky tests
+- Variable, method, and class names convey the subject
+- No console print statements
+- No unnecessary comments
+- New features have tests
+- Bugfixes have tests based on reproducers
+- StringBuilder or StringBuffer is used to perform multiple concatenations on strings (instead of '+')
```
akudiyar, line 12: This recommendation is in fact outdated. Since Java 7 StringBuilders are used internally for concatenations, so use them explicitly only when a string is built dynamically with different conditions,

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
...
+- For many choices `switch case` is used instead of multiple if-else conditions
```
akudiyar, line 13: I would also add a related recommendation for using enums over String constants. BTW, this and many other useful recommendations can be found in the brilliant book "Effective Java" by J. Bloch.

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
...
+- Used checked exceptions for recoverable operations and runtime exceptions for programming errors
```
akudiyar, line 14: We agreed for using RuntimeExceptions most of the time to not complicate the users' code and make it more usable in the reactive programming patterns. All internally produced exceptions must be classified and be instances of a TarantoolException class.

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
...
+- Code is DRY (see Do not Repeat Yourself principle)
```
akudiyar, line 15: I would relax this a bit by a rule "any simple code part shouldn't be repeated more than two times, and any complex code part should be enclosed in a corresponding method or a class". Obviously, complex code is constructed from simple code parts. You can think of a simple part like a one-liner.

docs/review-checklist.md
```diff
@@ -0,0 +1,20 @@
...
+- Code does not break SOLID principles
+- Methods grouped by functionality rather than by randomly, scope or accessibility. For example, a private class method
+  can be in between two public instance methods. The goal is to make reading and understanding the code easier.
+- Ensure override hashCode when overriding equals
```
akudiyar, line 19: I would suggest using generated hashCode/equals by your IDE (I suggest IDEA for this purpose). Don't write any custom code here if not strictly necessary. Also, don't add these methods if the class is not supposed to be used as a hash key or to be serialized. This will help to avoid the wrong usage.

docs/review-checklist.md
akudiyar: Also, most of the below recommendations and many more can be satisfied by an automatic code style fixing tool, like reformat command in IDEA or SonarLint in VStudio (https://code.visualstudio.com/docs/java/java-linting). Use reformatters and don't waste time on discussing code style changes, focus on things that matter instead.

akudiyar: One more thing that often irritates me -- code style changes in unrelated code parts (even in the same file) should go in separate PRs to not distract the readers from the actual changes. GH review interface is not very well in showing the in-line whitespace changes, so we'd better not mess up the PRs.

akudiyar (changes requested):

.github/pull_request_template.md
vrogach2020: Updating javadocs seems to be a part of release process and remain the responsibility of project maintainer.

docs/review-checklist.md
vrogach2020: > we need some limits for the PR size

Sane limit values are hard to imagine.
It's a 'hard to follow' and contradictory rule.

> recommendations for test coverage, javadocs

Not a subject of checklist. We must have simple and verifiable points here.
It can be a style guide or some coding conventions.

> below recommendations and many more can be satisfied by an automatic code style fixing tool

The developer is responsible for his code (it doesn't matter what tools he uses or not).
Automation of checking style is another problem/task. Here we just want to fix the aggreements on code style.

> code style changes in unrelated code parts 
As well as 1 - It's a 'hard to follow' and contradictory rule.

docs/review-checklist.md
vrogach2020: removed

docs/review-checklist.md
vrogach2020: Removed the point (it's not an ofter case). Good coding style can't be covered by this checklist

docs/review-checklist.md
vrogach2020: Removed a point. You're right but I don't want to metion TarantoolException here to avoid debates.

vrogach2020:
@akudiyar please review

docs/review-checklist.md
akudiyar: I think we can discuss these points later in DM

PR #209, opened 2022-04-21
Set correct fieldNumber in indexMetadata

### LINKED ISSUE #203: Fix primary key index definition error in update operation
The problem defining the index of a field with a primary key in function findSuitableIndex()
The problem is that fieldIndex = -1 in indexParts

```
`private static TarantoolIndexMetadata findSuitableIndex(TarantoolMetadataOperations operations,
                                                            TarantoolSpaceMetadata spaceMetadata,
                                                            Collection<TarantoolFieldMetadata> selectedFields) {
        Map<String, TarantoolIndexMetadata> allIndexes = operations.getSpaceIndexes(spaceMetadata.getSpaceName())
                .orElseThrow(() -> new TarantoolClientException(
                        "Metadata for space %s not found", spaceMetadata.getSpaceName()));

        TarantoolIndexMetadata suitableIndex = allIndexes.values().stream()
                .filter(metadata -> isSuitableIndex(metadata, selectedFields))
                .min(Comparator.comparingInt(m -> m.getIndexParts().size())) // Here
                .orElseThrow(() -> new TarantoolClientException("No indexes that fit the passed fields are found"));

        return suitableIndex;
    }`
```

dkasimovskiy:
There are two aspects of this issue:
1) Incorrect fieldNumber in the index part metadata. All fieldNumber decreased by one for all indexes.
2) When the driver try to find suitable index by field name it compares fieldNumber in the indexPartMetadata and selected fields. This can lead to a situation where the driver will select the wrong index or will not be able to find the index at all.

Closes #203

### DIFF
--- a/src/main/java/io/tarantool/driver/core/metadata/DDLTarantoolSpaceMetadataConverter.java
+++ b/src/main/java/io/tarantool/driver/core/metadata/DDLTarantoolSpaceMetadataConverter.java
@@ -20,8 +20,7 @@ import java.util.stream.Collectors;
 /**
  * Populates metadata from results of a call to proxy API function in Tarantool instance. The function result is
  * expected to have the format which is returned by DDL module.
- * See <a href="https://github.com/tarantool/ddl#input-data-format">
- * https://github.com/tarantool/ddl#input-data-format</a>
+ * See <a href="https://github.com/tarantool/ddl#input-data-format">https://github.com/tarantool/ddl#input-data-format</a>
  *
  * @author Sergey Volgin
  * @author Artyom Dubinin
@@ -233,7 +232,7 @@ public class DDLTarantoolSpaceMetadataConverter implements ValueConverter<Value,
                         }
                         String fieldType = fieldTypeValue.asStringValue().asString();
 
-                        return new TarantoolIndexPartMetadataImpl(fieldNumber - 1, fieldType, fieldPath);
+                        return new TarantoolIndexPartMetadataImpl(fieldNumber, fieldType, fieldPath);
                     })
                     .collect(Collectors.toList());
 
--- a/src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
+++ b/src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
@@ -68,6 +68,8 @@ public class ProxyTarantoolClientIT extends SharedCartridgeContainer {
     public static String PASSWORD;
 
     private static final String TEST_SPACE_NAME = "test__profile";
+    private static final String PK_FIELD_NAME = "profile_id";
+    private static final String BUCKET_ID_FIELD_NAME = "bucket_id";
     private static final int DEFAULT_TIMEOUT = 5 * 1000;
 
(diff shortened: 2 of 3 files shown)


src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
```diff
@@ -286,6 +288,29 @@ public void clusterUpdateTest() throws ExecutionException, InterruptedException
...
+    public void test_clusterUpdate_shouldReturnTuple_ifSearchByFieldName() throws ExecutionException,
+            InterruptedException {
+        TarantoolSpaceOperations<TarantoolTuple, TarantoolResult<TarantoolTuple>> profileSpace =
+                client.space("cursor_test_space");
+
+        int itemId = 223;
+        List<Object> rowData = Arrays.asList(itemId, "Jane Doe", 10, null);
+        TarantoolTuple tarantoolTuple = tupleFactory.create(rowData);
+
+        TarantoolResult<TarantoolTuple> insertTuples = profileSpace.insert(tarantoolTuple).get();
```
vrogach2020, line 301: Use replace() instead of insert(). Also please add the space 'cursor_test_space' to truncateSpace() setup block to avoid intersections.

src/main/java/io/tarantool/driver/core/metadata/DDLTarantoolSpaceMetadataConverter.java
```diff
@@ -233,7 +232,7 @@ private Map<String, TarantoolIndexMetadata> parseIndexes(Map<String, TarantoolFi
                         }
                         String fieldType = fieldTypeValue.asStringValue().asString();
 
-                        return new TarantoolIndexPartMetadataImpl(fieldNumber - 1, fieldType, fieldPath);
+                        return new TarantoolIndexPartMetadataImpl(fieldNumber, fieldType, fieldPath);
```
vrogach2020, line 235: Please add short description of the fix to changelog

vrogach2020 (changes requested):

src/main/java/io/tarantool/driver/core/metadata/DDLTarantoolSpaceMetadataConverter.java
dkasimovskiy: @ArtDu Nope. This is a bug. Possitions for indexes starts from 0.
Check with `box.space._vindex:select()`

vrogach2020 (approved):

src/main/java/io/tarantool/driver/core/metadata/DDLTarantoolSpaceMetadataConverter.java
ArtDu: @dkasimovskiy I checked ) 
```lua
tarantool> box.space._vindex:select()[1]
---
- [272, 0, 'primary', 'tree', {'unique': true}, [[0, 'string']]]
...

tarantool> box.space._vindex:select()[1][1]
---
- 272
...

tarantool>
```

PR #210, opened 2022-04-25
Use ValueType to search ValueConverters

### LINKED ISSUE #213: Use msgpack ValueType to search converters instead of class names
Because each type of Value has getValueType

```java
public interface Value
{
   
    ValueType getValueType();
...
}
```

Also javadoc of msgpack lib suggests to use this Enum ☺️

ArtDu:
Closes #213


src/main/java/io/tarantool/driver/mappers/DefaultMultiValueResultMapper.java
```diff
@@ -17,20 +18,6 @@
...
-     * @param valueConverter MessagePack value to object converter for each item
-     * @param contentClass target result content class
-     */
-    public DefaultMultiValueResultMapper(MessagePackMapper valueMapper,
-                                         ValueConverter<Value, T> valueConverter,
-                                         Class<T> contentClass) {
-        super(DefaultMessagePackMapperFactory.getInstance().emptyMapper(),
-                defaultValueConverter(valueConverter, contentClass), getResultClass(contentClass));
-    }
-
```
ArtDu, line 33: Weird unused code, we don't have an equivalent in `DefaultSingleValueResultMapper`

vrogach2020 (dismissed):
LGTM

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
```diff
@@ -40,65 +41,40 @@ public interface MessagePackMapperBuilder {
...
-     * @param <V>        MessagePack entity type
-     * @param <O>        object type
+     * @param <V>        MessagePack's entity type that the converter accepts and/or returns
+     * @param <O>        java object's type that the converter accepts and/or returns
      * @return builder
-     * @see io.tarantool.driver.mappers.DefaultMessagePackMapper#registerValueConverter(Class, ValueConverter)
+     * @see io.tarantool.driver.mappers.DefaultMessagePackMapper#registerValueConverter(ValueType, ValueConverter)
      */
     <V extends Value, O> MessagePackMapperBuilder
-    withValueConverter(Class<V> valueClass, ValueConverter<V, O> converter);
```
akudiyar, line 79: I basically don't like much the idea of **changing** the API for using the ValueType. As discussed, there may be cases when a Value class is more useful since it is a Java class and not a MessagePack type, and they don't have a  1-1 match.

I suggest adding a new interface with converters matched by ValueType and implement it, leaving the former variant as a backup. We can then `ValueType`s in the default  implementation for performance, but it will not break anything.

src/main/java/io/tarantool/driver/mappers/DefaultMultiValueResultMapper.java
akudiyar: The issue with this constructor is that it was intended to be used with the passed valueMapper, but it is possibly not safe.

src/main/java/io/tarantool/driver/mappers/DefaultMultiValueResultMapper.java
```diff
@@ -47,15 +32,6 @@ private static <T, R extends  List<T>> Class<MultiValueCallResult<T, R>> getResu
         return (Class<MultiValueCallResult<T, R>>) (Class<?>) MultiValueCallResult.class;
     }
 
-    private static
-    <T, R extends List<T>> ValueConverter<ArrayValue, ? extends MultiValueCallResult<T, R>>
-    defaultValueConverter(ValueConverter<Value, T> valueConverter,
```
akudiyar, line 52: Why this was removed?

akudiyar (changes requested):
The idea is not bad, but I suggest not changing the existing API that takes Java classes and just adding new methods. See comments below.

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
ArtDu: We still use the Value classes for all operations, except for one - finding converters in the Map. And there is no problem with it:

```
ImmutableNilValueImpl        -> NIL
ImmutableBooleanValueImpl    -> BOOLEAN

ImmutableBigIntegerValueImpl -> INTEGER
ImmutableLongValueImpl       -> INTEGER

ImmutableDoubleValueImpl     -> FLOAT
ImmutableStringValueImpl     -> STRING
ImmutableBinaryValueImpl     -> BINARY
ImmutableArrayValueImpl      -> ARRAY
ImmutableMapValueImpl        -> MAP
ImmutableExtensionValueImpl  -> EXTENSION
```
We use ValueType only in search, i.e. map certain converters to our ValueType and see no problem or benefit in using the class as the map key. On the contrary, it's the same approach as you used before with reflection when using [interfaces](https://github.com/tarantool/cartridge-java/blob/2bb9de7626b3a4004264e322047027f44f794fc9/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapperFactory.java#L31-L48)
```
ImmutableNilValueImpl        -> NilValue
ImmutableBooleanValueImpl    -> BooleanValue

ImmutableBigIntegerValueImpl -> IntegerValue
ImmutableLongValueImpl       -> IntegerValue

ImmutableDoubleValueImpl     -> FloatValue
ImmutableStringValueImpl     -> StringValue
ImmutableBinaryValueImpl     -> BinaryValue
ImmutableArrayValueImpl      -> ArrayValue
ImmutableMapValueImpl        -> MapValue
ImmutableExtensionValueImpl  -> ExtensionValue
```
This is just the same thing, but you don’t need to go to the root of the class. We just map the defined list of converters to some class whose root of that value is the ValueType. At the stage of comparison, we do not lose sight of any converter, and then you can always use the **canConvertValue** function of the converter and check exactly the class, if you somehow need it. Those, for example you have two converters for `ValueType.INTEGER`. These are BigIntegerConverter and LongConverter and inside you can already do something like this
```java
public class BigIntegerConverter implements ValueConverter<IntegerValue, BigInteger> {
    ...
    @Override
    public boolean canConvertValue(IntegerValue value) {
        return value.getClass().equals(ImmutableBigIntegerValueImpl.class);
    }
}
```

src/main/java/io/tarantool/driver/mappers/DefaultMultiValueResultMapper.java
ArtDu: We passed `valueMapper`, but it doesn't used
image

src/main/java/io/tarantool/driver/mappers/DefaultMultiValueResultMapper.java
ArtDu: This is internal code that is only used in the removed constructor of the `DefaultMultiValueResultMapper`. See above

src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
vrogach2020: @akudiyar @ArtDu I don't see any more blockers here for release. Artem done a great job simlifying the code. 
If we have any strange problem cases we'll fix it in later releases.

vrogach2020 (approved):

PR #215, opened 2022-05-01
Fix checkConverterByTargetType logic

### LINKED ISSUE #212: Remove getInterfaceParameterClass for checkConverterByTargetType
```java
/**
* Check if the specified converter can convert to the specified object type
*/
private boolean checkConverterByTargetType(ValueConverter<? extends Value, ?> converter, Class<?> targetClass) {
    try {
        ValueConverter<? extends Value, ?> exactMatch = valueConvertersByTarget.get(targetClass.getTypeName());
        return exactMatch == converter ||
                getInterfaceParameterClass(converter, converter.getClass(), 1).isAssignableFrom(targetClass);
    } catch (InterfaceParameterClassNotFoundException | InterfaceParameterTypeNotFoundException e) {
        return false;
    }
}
```

It's too slow. We must call this once when we register converter and cache that result in map for example
```java
private final Map<ValueConverter<? extends Value, ?>, String> targetClassesByValueConverters;
```

ArtDu:
Now the connector doesn't search converter's targetType by reflection,
the targetType is stored with the converter. Operations to convert a value with targetType are faster.

You can see result of perfomance tests here:
[master](https://github.com/tarantool/cartridge-java/runs/6320224857?check_suite_focus=true#step:7:1496)
[fix/check_converter_by_target_type](https://github.com/tarantool/cartridge-java/runs/6387468617?check_suite_focus=true#step:7:1975)

I didn't forget about

- [x] Tests (It isn't a new feature, all tests are the same)
- [x] Changelog
- [x] Documentation (It's iternal)

Closes #212

### DIFF
--- a/src/main/java/io/tarantool/driver/mappers/ConverterWrapper.java
+++ b/src/main/java/io/tarantool/driver/mappers/ConverterWrapper.java
@@ -1,0 +1,37 @@
+package io.tarantool.driver.mappers;
+
+import io.tarantool.driver.mappers.converters.Converter;
+import io.tarantool.driver.mappers.converters.ValueConverter;
+import io.tarantool.driver.mappers.converters.ObjectConverter;
+
+import java.io.Serializable;
+
+/**
+ * Base class for the internal logic of search converters.
+ * It is needed for faster possibility of obtaining a target type of converter.
+ * Target type is the second type parameter in {@link ValueConverter} or {@link ObjectConverter}.
+ *
+ * @param <C> the converter that returns an instance of this target type from an instance of input type
+ * @author Artyom Dubinin
+ */
+class ConverterWrapper<C extends Converter> implements Serializable {
+
+    private static final long serialVersionUID = 20220501L;
+
+    private final C converter;
+    private final Class<?> targetClass;
+
+    ConverterWrapper(C converter, Class<?> targetClass) {
+        this.converter = converter;
+        this.targetClass = targetClass;
+    }
+
+
+    public C getConverter() {
+        return converter;
+    }
+
+    public Class<?> getTargetClass() {
+        return targetClass;
+    }
+}
--- a/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
+++ b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
@@ -35,10 +35,8 @@ public class DefaultMessagePackMapper implements MessagePackMapper {
 
     private static final long serialVersionUID = 20220418L;
 
-    private final Map<ValueType, List<ValueConverter<? extends Value, ?>>> valueConverters;
-    private final Map<String, List<ObjectConverter<?, ? extends Value>>> objectConverters;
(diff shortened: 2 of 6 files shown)


src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
```diff
@@ -188,37 +182,16 @@ public <V extends Value, O> void registerValueConverter(Class<? extends V> value
...
                 valueConverters.computeIfAbsent(valueClass.getTypeName(), k -> new LinkedList<>());
-        converters.add(0, converter);
-        valueConvertersByTarget.put(objectClass.getTypeName(), converter);
+        converters.add(0, new ConverterWrapper<>(converter, objectClass));
     }
 
     /**
      * Check if the specified converter can convert to the specified object type
      */
-    private boolean checkConverterByTargetType(ValueConverter<? extends Value, ?> converter, Class<?> targetClass) {
```
akudiyar, line 200: Look, this logic allows to use the closest existing converter if a target converter doesn't exist in the mappers list. Don't remove this logic since it will break the SpringData internal object conversion algorithms, just make the reflection search an alternative way for the cases when the converter is not found by class directly in the list.

akudiyar: Of course, we will need tests for both cases.

akudiyar (changes requested):

src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
ArtDu: We have two searchs here which connected with an OR condition:
1. Search of the last registered converter for targetType in the `valueConvertersByTarget` map, which is a kind of cache since if you remove it, nothing should break. Please provide an example if this is not true.
2. Search targetType in the converter, in this place _\*Converter implements ValueConverter<\*, **targetType**>_ . So I just cached this search, but forgot to write `isAssignableFrom` instead of `equals`, this is my only mistake, as it seems to me. If I add `isAssignableFrom` instead of `equals` the same deterministic results as before

ArtDu: Rewrited to 
```java
return targetClassOfConverter.isAssignableFrom(targetClass);
```
Also I've tested this snapshot with cartridge-springdata and I haven't gotten any problems with it

wey1and (changes requested):

src/main/java/io/tarantool/driver/mappers/ConverterWrapper.java
```diff
@@ -0,0 +1,27 @@
+package io.tarantool.driver.mappers;
+
+import io.tarantool.driver.mappers.converters.Converter;
+
+import java.io.Serializable;
+
+class ConverterWrapper<C extends Converter> implements Serializable {
```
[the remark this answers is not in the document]
ArtDu: Fixed

wey1and (approved):

vrogach2020 (approved):

PR #216, opened 2022-05-04
Add workflow for performance tests in CI

ArtDu:
Now we can run these tests on a trigger in UI or via github rest.
Fixed unstoppable BenchmarkRunner.
Also added a manuall trigger in the default workflow.

You can see the result here https://github.com/tarantool/cartridge-java/actions/runs/2280786760

I didn't forget about

- [x] Tests (It's ci)
- [x] Changelog 
- [x] Documentation (it's iternal changes)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,7 @@
 # This workflow will build a Java project with Maven
 # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
 
-name: ubuntu-master
+name: Tests
 
 on:
   push:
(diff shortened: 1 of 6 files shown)


.github/workflows/tests.yml
```diff
@@ -1,9 +1,10 @@
 # This workflow will build a Java project with Maven
 # For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
 
-name: ubuntu-master
+name: Tests
 
 on:
+  workflow_dispatch:
```
dkasimovskiy, line 7: Why do we need to run tests manually?

dkasimovskiy (changes requested):

.github/workflows/tests.yml
ArtDu: It's not necessary, but i thought why not? 😁

dkasimovskiy: If it's not nessasary we no need it here.

ArtDu: Removed

dkasimovskiy (approved):

PR #228, opened 2022-05-24
WIP: Add varbinary tests

ArtDu:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Close #???


src/test/java/io/tarantool/driver/TarantoolUtils.java
```diff
@@ -0,0 +1,15 @@
+package io.tarantool.driver;
+
+public final class TarantoolUtils {
+    private TarantoolUtils() {
+    }
+
+    public static boolean versionGreaterThen(String version) {
+        return java.lang.System.getenv("TARANTOOL_VERSION") == null ||
```
wey1and, line 8: Possible NPE

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultStringValueToByteArrayConverter.java
```diff
@@ -0,0 +1,19 @@
...
+
+import io.tarantool.driver.mappers.converters.ValueConverter;
+import org.msgpack.value.StringValue;
+
+/**
+ * Default {@link StringValue} to {@code byte[]} converter
+ *
+ * @author Artyom Dubinin
+ */
+public class DefaultStringValueToByteArrayConverter implements ValueConverter<StringValue, byte[]> {
```
wey1and, line 11: I think we need implement canConvertValue
And add unit tests for these methods

wey1and (changes requested):

src/test/java/io/tarantool/driver/TarantoolUtils.java
ArtDu: Where exactly? We check that `java.lang.System.getenv("TARANTOOL_VERSION")` may be null in first condition

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultStringValueToByteArrayConverter.java
ArtDu: Unit test was added. We don't have to implement canConvertValue because StringValue and BinaryValue extend from the same superclass that uses byte array to store incoming messagepack value. Internally, StringValue and BinaryValue are the same.
image

wey1and (approved):

PR #233, opened 2022-06-02
Added more tests for SSL/TLS

### LINKED ISSUE #232: Add more tests for SSL/TLS
We need tests for mTLS and GOST cypher. These are specific cases. We need to be sure that they are supported by driver. Since the rest should work and have been tested in netty and OpenSSL libraries.

wey1and:
Added tests for mTLS and GOST cipher.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Close #232


pom.xml
```diff
@@ -220,6 +220,11 @@
             <version>5.7.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
akudiyar, line 223: I suggest making a new pom configuration (a child project) for GOST/CryptoPro support and adding new artifacts to Maven Central -- a "parent" project (may be the same) and the "with GOST" and "without GOST" versions of the driver. Let's discuss it verbally with the team.

akudiyar (changes requested):
GOST support should IMHO be separate from the main driver repository and moreover, it looks like being supported only with the Tarnatool EE. Let's discuss it may look like.

PR #247, opened 2022-08-07
Corrections based on the results of static analysis

dk2k:
Corrections based on the results of static analysis. Added a TODO to clarify if the fall through is switch () is intentional

### DIFF
--- a/src/main/java/io/tarantool/driver/codecs/MessagePackFrameDecoder.java
+++ b/src/main/java/io/tarantool/driver/codecs/MessagePackFrameDecoder.java
@@ -38,6 +38,7 @@ public class MessagePackFrameDecoder extends ReplayingDecoder<MessagePackFrameDe
                         checkpoint(DecoderState.BODY);
                     }
                     lenBuf.release();
+                    // TODO no break; and no comment if the fall through is intentional
                 case BODY:
                     if (size > 0) {
                         if (byteBuf.readableBytes() < size) {
--- a/src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
+++ b/src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
@@ -145,7 +145,4 @@ public class ClusterTarantoolTupleClient
         return new TarantoolTupleSpace(this, config, connectionManager, metadata, spaceMetadata);
     }
 
-    protected TarantoolClusterAddressProvider getAddressProvider() {
-        return super.getAddressProvider();
-    }
 }
--- a/src/main/java/io/tarantool/driver/handlers/TarantoolResponseHandler.java
+++ b/src/main/java/io/tarantool/driver/handlers/TarantoolResponseHandler.java
@@ -52,6 +52,9 @@ public class TarantoolResponseHandler extends SimpleChannelInboundHandler<Tarant
                         } catch (Throwable e) {
                             requestFuture.completeExceptionally(e);
                         }
+                    default:
+                        throw new IllegalStateException("unexpected response type " +
+                                tarantoolResponse.getResponseType());
                 }
             }
         } else {
--- a/src/main/java/io/tarantool/driver/protocol/TarantoolHeader.java
+++ b/src/main/java/io/tarantool/driver/protocol/TarantoolHeader.java
@@ -118,19 +118,21 @@ public final class TarantoolHeader implements Packable {
     public static TarantoolHeader fromMessagePackValue(Value value) throws TarantoolProtocolException {
         if (!value.isMapValue()) {
             throw new TarantoolProtocolException("TarantoolHeader can be unpacked only from MP_MAP, received "
-                    + value.toString());
+                    + value);
         }
         Map<Value, Value> values = value.asMapValue().map();
         TarantoolHeader header = new TarantoolHeader();
-        for (Value key: values.keySet()) {
+        for (Map.Entry<Value, Value> entry : values.entrySet()) {
+            Value key = entry.getKey();
             if (!key.isIntegerValue()) {
                 throw new TarantoolProtocolException("TarantoolHeader keys must be of MP_INT type");
             }
-            Value field = values.get(key);
+            Value field = entry.getValue();
             if (!field.isNumberValue()) {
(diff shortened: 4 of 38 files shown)


bitgorbovsky (changes requested):

bitgorbovsky:
@dk2k hello! Please, review comments and questions :)

src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
```diff
@@ -145,7 +145,4 @@ public ClusterTarantoolTupleClient(TarantoolClientConfig config, TarantoolCluste
         return new TarantoolTupleSpace(this, config, connectionManager, metadata, spaceMetadata);
     }
 
-    protected TarantoolClusterAddressProvider getAddressProvider() {
```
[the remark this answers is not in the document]
dk2k: It adds no logic and in fact overrides a parent's method. If this method is removed, the parent's one will be called. But without additional stack frame

src/main/java/io/tarantool/driver/protocol/TarantoolHeader.java
```diff
@@ -149,6 +151,8 @@ public static TarantoolHeader fromMessagePackValue(Value value) throws Tarantool
                 case IPROTO_TIMESTAMP:
                     header.setTimestamp(field.asFloatValue().toDouble());
                     break;
+                default:
```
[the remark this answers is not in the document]
dk2k: I know a lot about backward compatibility. Normally it's taken into account when you change the signature of a method - number and type of args. In this case you add a new method if it was public and preserve the old one. In this particular case the signature is left intact - no need to report a Runtime exception.
As a user I'd rather get an exception with some test when I pass an unexpected value. If I pass some value and rely on the fact that it has no effect currently - I rely on undefined behaviour, it's a house on the sand.
If you still think, that throwing an exception is a problem, I will remove it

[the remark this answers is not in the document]
dk2k: A very similar default block with IllegalStateException I added into TarantoolResponseHandler. Can be a subject of review. Just in case.

dk2k:
Removed default block as per request. Please review this change and my comments

dk2k:
Do you think that my changes broke the unit test?

src/main/java/io/tarantool/driver/codecs/MessagePackFrameDecoder.java
```diff
@@ -38,6 +38,7 @@ protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteB
                         checkpoint(DecoderState.BODY);
                     }
                     lenBuf.release();
+                    // TODO no break; and no comment if the fall through is intentional
```
bitgorbovsky, line 41: Yes, there shouldn't be a TODO. After reading IProto header with length of packet we can read body further without unnecessary break/return and swictj, just try to read available bytes.

bitgorbovsky (changes requested):
Sorry for the delay, I was on ill.
I see, that test for enterprise edition failed, we will see, what's wrong.

Seems, that in common this PR looks OK. Please, remove TODO (with `no break` comment).

src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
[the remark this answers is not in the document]
bitgorbovsky: Seems correct. OK

src/main/java/io/tarantool/driver/protocol/TarantoolHeader.java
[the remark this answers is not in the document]
bitgorbovsky: I'm not talking about backward compatibility in code, e.g between parent and child classes, different version of interfaces and so on. I'm talking about compatibility on a network protocol level (on IProto).

For example, you have service with cartidge-java 0.8.x (with `break` in this handler) working perfectly with Tarantool 2.8.x. Someday you decide to upgrade Tarantool to new experimental version and suddenly your service begin to fail.  Why? Because in experimental version some values added to response headers, and those value are unknown to current version of your java-driver. You need to upgrade you service too, rebuild, redeploy, etc...

So, to my mind, it's better to skip those values and continue response processing instead of throwing exceptions.

Extending protocol without changing previous semantic should not lead to failures.

src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
akudiyar, line 148: Removing this method may break the dependent libraries (e.g. java-replica). Please check twice that this method is not used, and first deprecate it.

src/main/java/io/tarantool/driver/handlers/TarantoolResponseHandler.java
```diff
@@ -52,6 +52,9 @@ protected void channelRead0(ChannelHandlerContext ctx, TarantoolResponse taranto
                         } catch (Throwable e) {
                             requestFuture.completeExceptionally(e);
                         }
+                    default:
+                        throw new IllegalStateException("unexpected response type " +
```
akudiyar, line 56: I would log a warning rather than fail with an exception. Throwing an exception here breaks the Open/Closed principle against IProto.

akudiyar (changes requested):
Thanks for the patch! Please take a look at the couple of comments below.

src/main/java/io/tarantool/driver/core/ClusterTarantoolTupleClient.java
dk2k: Restored the method

src/main/java/io/tarantool/driver/codecs/MessagePackFrameDecoder.java
dk2k: commented: fall through is intentional

src/main/java/io/tarantool/driver/handlers/TarantoolResponseHandler.java
dk2k: changed to log message

akudiyar (approved):
LGTM

PR #255, opened 2022-09-12
Add 'fields' parameter to ProxySpace to Conditions

### LINKED ISSUE #236: Implement API to pass 'fields' parameter to CRUD library
В java любая crud операция возвращает всегда все поля (весь tuple целиком). в crud есть возможность задать в opts столбцы (fields), которые нужно вернуть, но в cartridge-driver 0.7.2 это задать нельзя, opts там хардкодится, единственное, что ставится, это таймаут.  нельзя ли ликвидировать "разрыв" в функциональности между java и crud ?

к примеру, в кортеже 30 полей, и я не могу выбрать одно поле, только все 30. 
имеется в виду:

space = tntClient.getClient().space(SPACE_NAME);
space.select(Conditions.any()) — возвращает все поля, нет возможности задать нужное (ые), condition для примера

тоже  самое replace/update/upsert/delete

гоняют по сети впустую данные

ArtDu:

I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

Closes #236

### DIFF
--- a/src/main/java/io/tarantool/driver/core/proxy/AbstractProxyOperation.java
+++ b/src/main/java/io/tarantool/driver/core/proxy/AbstractProxyOperation.java
@@ -13,6 +13,7 @@ import java.util.concurrent.CompletableFuture;
  *
  * @author Alexey Kuzin
  * @author Sergey Volgin
+ * @author Artyom Dubinin
  */
 abstract class AbstractProxyOperation<T> implements ProxyOperation<T> {
 
@@ -54,5 +55,69 @@ abstract class AbstractProxyOperation<T> implements ProxyOperation<T> {
     @Override
     public CompletableFuture<T> execute() {
         return client.callForSingleResult(functionName, arguments, argumentsMapper, resultMapper);
+    }
+
+    abstract static class GenericOperationsBuilder<T, B extends GenericOperationsBuilder<T, B>> {
+        protected TarantoolCallOperations client;
+        protected String spaceName;
+        protected String functionName;
+        protected MessagePackObjectMapper argumentsMapper;
+        protected CallResultMapper<T, SingleValueCallResult<T>> resultMapper;
+        protected int requestTimeout;
+
+        GenericOperationsBuilder() {
+        }
+
+        abstract B self();
+
+        /**
+         * Specify a client for sending and receiving requests from Tarantool server
+         * @param client Tarantool server client
+         * @return builder
+         */
+        public B withClient(TarantoolCallOperations client) {
+            this.client = client;
+            return self();
+        }
+
+        /**
+         * Specify name of Tarantool server space to work with
+         * @param spaceName name of Tarantool server space
+         * @return builder
+         */
+        public B withSpaceName(String spaceName) {
+            this.spaceName = spaceName;
+            return self();
+        }
+
+        /**
+         * Specify name of the Tarantool server function called through preparing request
+         * @param functionName name of Tarantool server function
+         * @return builder
+         */
+        public B withFunctionName(String functionName) {
+            this.functionName = functionName;
+            return self();
+        }
+
+        public B withArgumentsMapper(MessagePackObjectMapper objectMapper) {
+            this.argumentsMapper = objectMapper;
+            return self();
+        }
+
+        public B withResultMapper(CallResultMapper<T, SingleValueCallResult<T>> resultMapper) {
+            this.resultMapper = resultMapper;
+            return self();
+        }
+
+        /**
+         * Specify response reading timeout
+         * @param requestTimeout the timeout for reading the responses from Tarantool server, in milliseconds
+         * @return builder
+         */
+        public B withRequestTimeout(int requestTimeout) {
+            this.requestTimeout = requestTimeout;
+            return self();
+        }
     }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java
```diff
@@ -207,12 +245,12 @@ private CompletableFuture<R> upsert(Conditions conditions,
     public CompletableFuture<Void> truncate() throws TarantoolClientException {
         try {
             return executeVoidOperation(TruncateProxyOperation.<Void>builder()
-                    .withClient(client)
-                    .withSpaceName(spaceName)
-                    .withFunctionName(operationsMapping.getTruncateFunctionName())
-                    .withRequestTimeout(config.getRequestTimeout())
-                    .build()
-            );
+                                                .withClient(client)
```
akudiyar, line 248: Extra spacing?

src/main/java/io/tarantool/driver/core/proxy/AbstractProxyOperation.java
```diff
@@ -55,4 +56,84 @@ public CallResultMapper<T, SingleValueCallResult<T>> getResultMapper() {
...
+         *
+         * @param functionName name of Tarantool server function
+         * @return builder
+         */
+        public B withFunctionName(String functionName) {
+            this.functionName = functionName;
+            return self();
+        }
+
+        public B withArgumentsMapper(MessagePackObjectMapper objectMapper) {
```
akudiyar, line 107: These two methods are missing a Javadoc

akudiyar (changes requested):
I'm not sure about putting `fields` into `Conditions`. This class was intended and can be used only for select operations. Using it for the other operations that cannot make use of other `Conditions` fields leads to user confusion and we will have to add a lot of extra checks that the user haven't specified any unsupported fields.

 I see two options here:
 - making a new class that is used as base class for Conditions and can be passed to all operations, so it does not allow using extra fields where they aren't supported
 - judging by the output type about what fields should be included. This approach can be combined with the first one or implemented independently. But to me it looks closer to what users may expect and eliminate any confusion that may arise from the difference between the passed field names and the available output type fields. However, matching fields to the output type may be handed over to the Spring Data connector, as it has more control over the output structures.
 
 BTW, I like the ProxyOperations refactoring, wouldn't it be better to merge it separately first?

src/main/java/io/tarantool/driver/core/proxy/AbstractProxyOperation.java
ArtDu: Fixed and exported to https://github.com/tarantool/cartridge-java/pull/256

ArtDu:
> I'm not sure about putting `fields` into `Conditions`. This class was intended and can be used only for select operations. Using it for the other operations that cannot make use of other `Conditions` fields leads to user confusion and we will have to add a lot of extra checks that the user haven't specified any unsupported fields.
> 
> I see two options here:
> 

> * judging by the output type about what fields should be included. This approach can be combined with the first one or implemented independently. But to me it looks closer to what users may expect and eliminate any confusion that may arise from the difference between the passed field names and the available output type fields. However, matching fields to the output type may be handed over to the Spring Data connector, as it has more control over the output structures.
> 
> BTW, I like the ProxyOperations refactoring, wouldn't it be better to merge it separately first?

I agree with you that using `Condition` is not the best solution. It's interesting about seeing in output type, but we need a more general way to pass not only `fields` but other parameters in crud(it's possible we add this option later). Maybe we add another class - `Option` and will inherit it for specific methods options? I don't see how to use `Condition` and base parent class, because we will still have all `select` options in Condition and it will be confusing

akudiyar:
The fact that any other methods are using Conditions now even if they aren't able to use all options there needs to be fixed, I think.
An option that I see here -- to add an `OperationOptions` base class, that will contain base options like timeouts, pass it to every method, and for CRUD/Proxy space/client implementation allow to pass a child class there that supports more options, only specific to the cluster client.
That means we will leave the existing method signatures as is, and add new variations supporting this new `OperationOption` as a parameter. For the single node implementation, it will support only the base set of options (like timeouts only). Thus we will not break the single server compatibility.

Examples:
 - Single node client:
```java
   Conditions conditions = Conditions.any();
   OperationOptions options = BaseOperationOptions.builder().withTimeout(100500).build();
   client.space("test").select(conditions, options);
``` 
 - Cluster client:
 ```java
   Conditions conditions = Conditions.any();
   OperationOptions options = BaseOperationOptions.builder()
       .withTimeout(100500)
       .withFields(Arrays.asList("id", "name"))
       .build();
   client.space("test").select(conditions, options);
 ```
 
 Notes:
  - we'll have to have casts in the cluster space implementation in the naive solution. The other (better) option is to parameterize the space and client classes with the options class, sigh... (I hate Java)
  - we'll have to generate another dozen of methods with parameter combinations for this new parameter for better UX (I hate Java).
  - at the moment we are not passing the `batchSize` option correctly to tarantool/crud because we don't have the same option in the standalone server (we pass the `Conditions.limit()` value to that). It will be better to have this option in the new class. However, it's not right for other options like `startAfter`, because they are related to the select operation semantics and not to the internal tarantool/crud library tweaks.
  - the new options related to batch operations will also be supported by this new options class.

PR #258, opened 2022-09-13
Refactor CRUDOperationOptions to a hierarchy of classes

akudiyar:
- Make CRUD operation options more agile for further adding support of new tarantool/crud features.
 - Add fluent builders for options classes

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation


src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * This class is not part of the public API.
+ *
+ * Represent basic options for cluster operations
+ *
+ * @author Alexey Kuzin
+ */
+public class CRUDBaseOperationOptions extends CRUDAbstractOperationOptions {
+
+    public static final String TIMEOUT = "timeout";
```
ArtDu, line 15: Let's add here `vshard_router` option. Used for every operation in vshard except special ones (cut_rows, cut_objects, stats)

ArtDu (changes requested):
I've been trying merge two Builders in class and came to the conclusion that this is probably the best way. But we need to add javadocs to Builders because it's too complicated and may misunderstanding or I've never seen this pattern before

Then after merging your PR I will add public methods for ProxyTarantoolSpace operations with CRUD*OperationOptions. 'Cause I think it's the only way we can follow.

Also don't forget to add few words to the Changelog, I know that is internal, but I'm thinking all changes are important

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Map;
+
+/**
+ * This class is not part of the public API.
+ *
+ * Represent basic options for cluster operations
+ *
+ * @author Alexey Kuzin
+ */
+public class CRUDBaseOperationOptions extends CRUDAbstractOperationOptions {
```
ArtDu, line 13: Perhaps rename to CrudBaseOperationOptions? Because it reads like `CRUDB`

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
```diff
@@ -0,0 +1,49 @@
+package io.tarantool.driver.core.proxy;
+
+import java.util.HashMap;
+import java.util.Map;
```
ArtDu, line 4: Unused import statement

src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOperationOptions.java
```diff
@@ -0,0 +1,74 @@
+package io.tarantool.driver.core.proxy;
+
+import io.tarantool.driver.protocol.Packable;
+import java.util.Map;
```
ArtDu, line 4: Unused import statement

src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOperationOptions.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * This class is not part of the public API.
+ *
+ * Represent options for proxy cluster select operations
+ *
+ * @author Sergey Volgin
+ * @author Alexey Kuzin
+ */
+public final class CRUDSelectOperationOptions extends CRUDBaseOperationOptions {
```
ArtDu, line 14: Maybe we'll reduce class name to `CrudSelectOptions`, obviously that select is operation

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
akudiyar: I don't want to add any new things to this PR. They will follow in separate PRs - we should keep the changesets as smaller as possible. Refactorings are better to be done separately from the new features.

akudiyar: Created #261 for that

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
akudiyar: We already have this prefix for other parts + CRUD is both an abbreviature and a library name. I'd rather use any other name as a prefix if you have suggestions. But  we need to maintain the dependency with tarantool/crud library now (until we refactor everything to support different communication bridges).

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
ArtDu: Yeah, you are right, we could add this feature if it needed later

src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOperationOptions.java
akudiyar: Renamed derived classes

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOperationOptions.java
akudiyar: Fixed

src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOperationOptions.java
akudiyar: Fixed

akudiyar:
I decided to not add the changelog now because I'm going to push another PR immediately with the batch options and there I can add the changelog entry for all changes.

akudiyar:
NB: we cannot use these CRUD*Options classes for representing the user-defined operation options. These are internal containers for passing the options and they include the values from Conditions, for example. These classes are internal and shouldn't leak to API.

For the API, we will have to prepare another hierarchy of options classes, with different available fields and obviously without CRUD prefix.

ArtDu:
> NB: we cannot use these CRUD*Options classes for representing the user-defined operation options. These are internal containers for passing the options to tarantool/crud API and they include the values from Conditions, for example. These classes are internal and shouldn't leak to API.
> 
> For the API, we will have to prepare another hierarchy of options classes, with different available fields and obviously without CRUD prefix.

Yep, and passing in `Conditions` `startTuple`, `offset` and `limit` was a wrong idea. Because we have offset and startTuple options that are unique for each of the functions. Also we have problem with the `Condition` that we use it not only in `select`, but for update, upsert and etc. which ones don't use these options and that's weird.

I believe we should have implemented API that as close as possible to the original API and futher will add easy to use constructions.

But now we have what we have. And I agree with you that's we need another Options hierarchy so as not to multiply collisions but we will have exactly same structure like CrudBase*Options except options from `Conditions` and it's hard for eyes

ArtDu (approved):

src/main/java/io/tarantool/driver/core/proxy/CRUDAbstractOperationOptions.java
```diff
@@ -0,0 +1,53 @@
...
+ * pattern for the cluster proxy operation options builders.
+ *
+ * @author Alexey Kuzin
+ */
+abstract class CRUDAbstractOperationOptions {
+
+    private final Map<String, Object> resultMap = new HashMap<>();
+
+    /**
+     * Inheritable Builder for select cluster proxy operation options.
```
ArtDu, line 19: I think not only `select cluster proxy operations` we can use the class for modified options too

src/main/java/io/tarantool/driver/core/proxy/CRUDAbstractOperationOptions.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * This class is not part of the public API.
+ *
+ * An abstract class necessary for implementing CRT (curiously recurring template)
+ * pattern for the cluster proxy operation options builders.
+ *
+ * @author Alexey Kuzin
+ */
+abstract class CRUDAbstractOperationOptions {
```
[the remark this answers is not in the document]
akudiyar: Let's think about it, but not in that PR. I doubt that we should use a class from internals in the API.

src/main/java/io/tarantool/driver/core/proxy/CRUDAbstractOperationOptions.java
akudiyar: Fixed in #264

PR #263, opened 2022-09-15
Add 'fields' parameter to ProxySpace

ArtDu:
Add `fields` option to crud proxy methods

- [x] Tests
- [x] Changelog
- [ ] Documentation

Closes https://github.com/tarantool/cartridge-java/issues/236

### DIFF
--- a/src/main/java/io/tarantool/driver/api/space/options/OperationWithFieldsOptions.java
+++ b/src/main/java/io/tarantool/driver/api/space/options/OperationWithFieldsOptions.java
@@ -1,0 +1,36 @@
+package io.tarantool.driver.api.space.options;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Base interface for all operation options that may have a configurable return.
+ *
+ * @author Artyom Dubinin
+ */
+public interface OperationWithFieldsOptions<T extends OperationWithFieldsOptions<T>>
+    extends Options, Self<T> {
+
+    String FIELDS = "fields";
+
+    /**
+     * Specifies list of fields names for getting only a subset of fields.
+     * By default, all fields are returned.
+     *
+     * @param fields list of string field names
+     * @return this options instance
+     */
+    default T withFields(List<String> fields) {
+        addOption(FIELDS, fields);
+        return self();
+    }
+
+    /**
+     * Return list of fields names for getting only a subset of fields.
+     *
+     * @return list of fields string names
+     */
+    default Optional<List> getFields() {
+        return getOption(FIELDS, List.class);
+    }
+}
--- a/src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
+++ b/src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
@@ -24,12 +24,11 @@ class CRUDBaseOptions extends CRUDAbstractOperationOptions {
      *
      * @see CRUDAbstractOperationOptions.AbstractBuilder
      */
-    protected abstract static
-    class AbstractBuilder<O extends CRUDBaseOptions, T extends AbstractBuilder<O, T>>
-        extends CRUDAbstractOperationOptions.AbstractBuilder<O, T> {
+    protected abstract static class AbstractBuilder<O extends CRUDBaseOptions, B extends AbstractBuilder<O, B>>
+        extends CRUDAbstractOperationOptions.AbstractBuilder<O, B> {
         protected Optional<Integer> timeout = Optional.empty();
 
-        public T withTimeout(Optional<Integer> timeout) {
+        public B withTimeout(Optional<Integer> timeout) {
             this.timeout = timeout;
             return self();
         }
--- a/src/main/java/io/tarantool/driver/core/proxy/CRUDBatchOptions.java
+++ b/src/main/java/io/tarantool/driver/core/proxy/CRUDBatchOptions.java
@@ -9,7 +9,7 @@ import java.util.Optional;
  *
  * @author Alexey Kuzin
  */
-public final class CRUDBatchOptions extends CRUDBaseOptions {
+final class CRUDBatchOptions extends CRUDReturnOptions {
 
     public static final String BATCH_STOP_ON_ERROR = "stop_on_error";
     public static final String BATCH_ROLLBACK_ON_ERROR = "rollback_on_error";
@@ -17,30 +17,27 @@ public final class CRUDBatchOptions extends CRUDBaseOptions {
     private <T extends AbstractBuilder<T>>
     CRUDBatchOptions(AbstractBuilder<T> builder) {
         super(builder);
-
         addOption(BATCH_STOP_ON_ERROR, builder.stopOnError);
         addOption(BATCH_ROLLBACK_ON_ERROR, builder.rollbackOnError);
     }
 
-    protected abstract static
-    class AbstractBuilder<T extends AbstractBuilder<T>>
-        extends CRUDBaseOptions.AbstractBuilder<CRUDBatchOptions, T> {
+    protected abstract static class AbstractBuilder<B extends AbstractBuilder<B>>
+            extends CRUDReturnOptions.AbstractBuilder<CRUDBatchOptions, B> {
         private Optional<Boolean> stopOnError = Optional.empty();
         private Optional<Boolean> rollbackOnError = Optional.empty();
 
-        public T withStopOnError(Optional<Boolean> stopOnError) {
+        public B withStopOnError(Optional<Boolean> stopOnError) {
             this.stopOnError = stopOnError;
             return self();
         }
 
-        public T withRollbackOnError(Optional<Boolean> rollbackOnError) {
+        public B withRollbackOnError(Optional<Boolean> rollbackOnError) {
             this.rollbackOnError = rollbackOnError;
             return self();
         }
     }
 
     protected static final class Builder extends AbstractBuilder<Builder> {
-
         @Override
         Builder self() {
             return this;
--- a/src/main/java/io/tarantool/driver/core/proxy/CRUDReturnOptions.java
+++ b/src/main/java/io/tarantool/driver/core/proxy/CRUDReturnOptions.java
@@ -1,0 +1,48 @@
+package io.tarantool.driver.core.proxy;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class is not part of the public API.
+ * <p>
+ * Represent returned options for cluster proxy operations.
+ * The return set of fields of the tuples can be specified.
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+class CRUDReturnOptions extends CRUDBaseOptions {
+
+    public static final String FIELDS = "fields";
+
+    protected <O extends CRUDReturnOptions, B extends AbstractBuilder<O, B>>
+    CRUDReturnOptions(CRUDReturnOptions.AbstractBuilder<O, B> builder) {
+        super(builder);
+        addOption(FIELDS, builder.fields);
+    }
+
+    protected abstract static
+    class AbstractBuilder<O extends CRUDReturnOptions, B extends AbstractBuilder<O, B>>
+            extends CRUDBaseOptions.AbstractBuilder<O, B> {
+        private Optional<List> fields = Optional.empty();
+
+        public B withFields(Optional<List> fields) {
+            this.fields = fields;
+            return self();
+        }
+    }
+
+    protected static final class Builder extends AbstractBuilder<CRUDReturnOptions, Builder> {
+
+        @Override
+        CRUDReturnOptions.Builder self() {
+            return this;
+        }
+
+        @Override
+        public CRUDReturnOptions build() {
+            return new CRUDReturnOptions(this);
+        }
+    }
+}
--- a/src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOptions.java
+++ b/src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOptions.java
@@ -2,6 +2,7 @@ package io.tarantool.driver.core.proxy;
 
 import io.tarantool.driver.protocol.Packable;
 
+import java.util.List;
 import java.util.Optional;
 
 /**
@@ -18,14 +19,15 @@ final class CRUDSelectOptions extends CRUDBucketIdOptions {
     public static final String SELECT_LIMIT = "first";
     public static final String SELECT_AFTER = "after";
     public static final String SELECT_BATCH_SIZE = "batch_size";
+    public static final String FIELDS = "fields";
 
-    private <T extends AbstractBuilder<T>>
-    CRUDSelectOptions(AbstractBuilder<T> builder) {
+    private <B extends AbstractBuilder<B>> CRUDSelectOptions(AbstractBuilder<B> builder) {
         super(builder);
 
         addOption(SELECT_LIMIT, builder.selectLimit);
         addOption(SELECT_AFTER, builder.after);
         addOption(SELECT_BATCH_SIZE, builder.selectBatchSize);
+        addOption(FIELDS, builder.fields);
     }
 
     /**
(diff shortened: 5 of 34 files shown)


src/test/java/io/tarantool/driver/integration/ProxyTarantoolSpaceIT.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Collections;
+import java.util.concurrent.ExecutionException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/**
+ * @author Artyom Dubinin
+ */
+public class ProxyTarantoolSpaceIT extends SharedCartridgeContainer {
```
akudiyar, line 41: Why do we have a new test class?

akudiyar (changes requested):
I liked it overall, but I believe that there's a better way. Please take a look at the comments below.

I think that `Options` should be much like the `Conditions` class, which is actually a builder and it is convenient to call a select method like `space.select(Conditions.any())`, for example. We should have something like that for `Options`. Of course, in this case, we'll have to tell the user if any actual options aren't supported by the particular client instance (in the same way as we do in Conditions). Serialization of `Options` instances will happen internally, no need to force users to use additional builder instances and call `build()`, that's ugly. But the difference will be that we can have different `Options` classes for each operation.  So for `select` it will be `SelectOptions`, for `insertMany` it will be `InsertManyOptions` etc.

An example of how the `Options` class can be used: 
```java
    space.select(
        Conditions.any(),
        SelectOptions.withTimeout(100500).withBatchSize(1000).withFields(Arrays.asList("id"))
    )
```
Here `SelectOptions` inherit some base class that provides `withTimeout` and `withFields` methods.

P.S. I think we further will replace the single `Conditions` class in the methods where it is passed with a hierarchy, like `PrimaryIndexConditions` (or just `PrimaryIndex`) for the methods that support only it and full `Conditions` (or `IndexConditions`?) for `select`.

ArtDu:
> An example of how the `Options` class can be used:
> 
> ```java
>     space.select(
>         Conditions.any(),
>         SelectOptions.withTimeout(100500).withBatchSize(1000).withFields(Arrays.asList("id"))
>     )
> ```
> 
> Here `SelectOptions` inherit some base class that provides `withTimeout` and `withFields` methods.

I see only one way how to implement this, we should have all abstract classes except leaf in inheritance hierarchy. Leaf implement `self` and also we should have one entrypoint in leaf, something like `create`. 

```java
SelectOptions.create().withTimeout(100500).withBatchSize(1000).withFields(Arrays.asList("id"))
```

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
```diff
@@ -28,11 +31,17 @@ public class CRUDBaseOptions extends CRUDAbstractOperationOptions {
...
         extends CRUDAbstractOperationOptions.AbstractBuilder<O, T> {
         protected Integer timeout;
+        private Options options;
 
         public T withTimeout(int timeout) {
             this.timeout = timeout;
             return self();
         }
+
+        public T withOptions(Options options) {
```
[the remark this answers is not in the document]
ArtDu: In merged PR `withTimeout` is using as default parameter(to pass config.timeout) and `withOptions -> timeout` as custom timeout

src/main/java/io/tarantool/driver/core/proxy/CRUDFunctionOptions.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * This class is not part of the public API.
+ *
+ * Represent options for cluster proxy operations which returns fields.
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+class CRUDFunctionOptions extends CRUDBaseOptions {
```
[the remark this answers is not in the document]
ArtDu: In merged [PR](https://github.com/tarantool/cartridge-java/pull/267) we've discussed to name `CRUDReturnOptions`

src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java
```diff
@@ -207,12 +218,12 @@ private CompletableFuture<R> upsert(Conditions conditions,
     public CompletableFuture<Void> truncate() throws TarantoolClientException {
         try {
             return executeVoidOperation(TruncateProxyOperation.<Void>builder()
-                    .withClient(client)
-                    .withSpaceName(spaceName)
-                    .withFunctionName(operationsMapping.getTruncateFunctionName())
-                    .withRequestTimeout(config.getRequestTimeout())
-                    .build()
-            );
+                                                .withClient(client)
```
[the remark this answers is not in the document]
ArtDu: Fixed in merged [PR](https://github.com/tarantool/cartridge-java/pull/267)

src/main/java/io/tarantool/driver/core/space/options/Options.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Public API for space operations.
+ *
+ * An abstract class necessary for implementing CRT (curiously recurring template)
+ * pattern for the cluster proxy operation options builders.
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+public class Options {
```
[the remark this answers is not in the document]
ArtDu: Fixed in merged [PR](https://github.com/tarantool/cartridge-java/pull/267)

src/test/java/io/tarantool/driver/integration/ProxyTarantoolSpaceIT.java
ArtDu: Added some classes in merged [PR](https://github.com/tarantool/cartridge-java/pull/267). I did this because the main class `ProxyAPI` is overloaded with many testcases, I wanted more precision test classes

src/main/java/io/tarantool/driver/api/space/options/OperationWithFieldsOptions.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Base interface for all operation options that may have a configurable return.
+ *
+ * @author Artyom Dubinin
+ */
+public interface OperationWithFieldsOptions extends OperationWithTimeoutOptions {
+    /**
+     * Return list of fields.
+     *
+     * @return timeout, in milliseconds.
```
akudiyar, line 15: Not timeout, fields

src/main/java/io/tarantool/driver/api/space/options/OperationWithFieldsOptions.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Optional;
+
+/**
+ * Base interface for all operation options that may have a configurable return.
+ *
+ * @author Artyom Dubinin
+ */
+public interface OperationWithFieldsOptions extends OperationWithTimeoutOptions {
+    /**
+     * Return list of fields.
```
akudiyar, line 13: It would be better to have a more full description (what is this list for?)

src/main/java/io/tarantool/driver/api/space/options/proxy/ProxyReturnOptions.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+abstract class ProxyReturnOptions<B extends ProxyReturnOptions<B>> extends ProxyBaseOptions<B>
+        implements OperationWithFieldsOptions {
+
+    public static final String FIELDS = "fields";
+
+    /** TODO:
```
akudiyar, line 20: Doc TODO?

src/main/java/io/tarantool/driver/api/space/options/proxy/ProxyReturnOptions.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param fields request timeout, in milliseconds
+     * @return this options instance
+     */
+    public B withFields(List<String> fields) {
+        addOption(FIELDS, fields);
+        return self();
+    }
+
+    @Override
+    public Optional<List> getFields() {
```
akudiyar, line 33: I think we need `List<String>` here

src/main/java/io/tarantool/driver/core/proxy/CRUDBatchOptions.java
```diff
@@ -9,38 +9,38 @@
...
-public final class CRUDBatchOptions extends CRUDBaseOptions {
+public final class CRUDBatchOptions extends CRUDReturnOptions {
 
     public static final String BATCH_STOP_ON_ERROR = "stop_on_error";
     public static final String BATCH_ROLLBACK_ON_ERROR = "rollback_on_error";
 
     protected
-    <T extends AbstractBuilder<T>>
-    CRUDBatchOptions(AbstractBuilder<T> builder) {
+    <O extends CRUDBatchOptions, B extends AbstractBuilder<O, B>>
```
akudiyar, line 18: `CRUDBatchOptions` is a final class, we don't need a generic for Options here

src/main/java/io/tarantool/driver/core/proxy/CRUDReturnOptions.java
```diff
@@ -0,0 +1,48 @@
+package io.tarantool.driver.core.proxy;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class is not part of the public API.
+ * <p>
+ * Represent options for cluster proxy operations which returns fields
```
akudiyar, line 9: Let's reformulate it. Something like `operations for that the set of fields returned in the result tuples can be specified`?

src/main/java/io/tarantool/driver/core/proxy/CRUDSelectOptions.java
```diff
@@ -13,14 +13,14 @@
...
-final class CRUDSelectOptions extends CRUDBaseOptions {
+final class CRUDSelectOptions extends CRUDReturnOptions {
 
     public static final String SELECT_LIMIT = "first";
     public static final String SELECT_AFTER = "after";
     public static final String SELECT_BATCH_SIZE = "batch_size";
 
-    private <T extends AbstractBuilder<T>>
-    CRUDSelectOptions(AbstractBuilder<T> builder) {
+    private <O extends CRUDSelectOptions, B extends AbstractBuilder<O, B>>
```
akudiyar, line 22: THis class is final, the generic for `Options` is redundant.

akudiyar (changes requested):
Almost good, the questions are mostly about the docstrings.

ArtDu:
> Almost good, the questions are mostly about the docstrings.

Yeah, that is reason why I've converted PR to Draft

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
```diff
@@ -15,8 +15,8 @@ class CRUDBaseOptions extends CRUDAbstractOperationOptions {
     public static final String TIMEOUT = "timeout";
 
     protected
-    <O extends CRUDBaseOptions, T extends AbstractBuilder<O, T>>
-    CRUDBaseOptions(AbstractBuilder<O, T> builder) {
+    <O extends CRUDBaseOptions, B extends AbstractBuilder<O, B>>
+    CRUDBaseOptions(AbstractBuilder<O, B> builder) {
```
Elishtar, line 19: I guess this case should be excluded from checkstyle only one line, current format looks not readable

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
```diff
@@ -26,11 +26,11 @@ class CRUDBaseOptions extends CRUDAbstractOperationOptions {
      * @see CRUDAbstractOperationOptions.AbstractBuilder
      */
     protected abstract static
-    class AbstractBuilder<O extends CRUDBaseOptions, T extends AbstractBuilder<O, T>>
-        extends CRUDAbstractOperationOptions.AbstractBuilder<O, T> {
+    class AbstractBuilder<O extends CRUDBaseOptions, B extends AbstractBuilder<O, B>>
+        extends CRUDAbstractOperationOptions.AbstractBuilder<O, B> {
```
Elishtar, line 30: I think. looks better
    protected abstract static class AbstractBuilder<O extends CRUDBaseOptions, B extends AbstractBuilder<O, B>> 
            extends CRUDAbstractOperationOptions.AbstractBuilder<O, B> {
        
        protected Optional<Integer> timeout = Optional.empty();

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
```diff
@@ -26,11 +26,11 @@ class CRUDBaseOptions extends CRUDAbstractOperationOptions {
...
         protected Optional<Integer> timeout = Optional.empty();
 
-        public T withTimeout(Optional<Integer> timeout) {
```
Elishtar, line 33: Just question why? T looks more standard

src/main/java/io/tarantool/driver/core/proxy/CRUDReturnOptions.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author Artyom Dubinin
+ */
+class CRUDReturnOptions extends CRUDBaseOptions {
+
+    public static final String FIELDS = "fields";
+
+    protected <O extends CRUDReturnOptions, B extends AbstractBuilder<O, B>>
+    CRUDReturnOptions(CRUDReturnOptions.AbstractBuilder<O, B> builder) {
+        super(builder);
+
```
Elishtar, line 21: Remove space

src/main/java/io/tarantool/driver/api/space/options/OperationWithFieldsOptions.java
ArtDu: Fixed

src/main/java/io/tarantool/driver/core/proxy/CRUDFunctionOptions.java
```diff
@@ -0,0 +1,65 @@
+package io.tarantool.driver.core.proxy;
+
+import io.tarantool.driver.core.space.options.Options;
+
+import java.util.Map;
+
+/**
+ * This class is not part of the public API.
+ *
+ * Represent options for cluster proxy operations which returns fields.
```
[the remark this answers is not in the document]
ArtDu: It has been rewritten

src/main/java/io/tarantool/driver/api/space/TarantoolSpaceOperations.java
```diff
@@ -47,14 +48,24 @@
...
      * @param conditions query with options
      * @return a future that will contain all corresponding tuples once completed
      * @throws TarantoolClientException in case if the request failed
      */
     CompletableFuture<R> select(Conditions conditions) throws TarantoolClientException;
 
+    /**
+     * Select tuples matching the specified query with specified conditions and options.
+     *
+     * @param conditions specified conditions
```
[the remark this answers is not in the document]
ArtDu: Fixed

src/main/java/io/tarantool/driver/api/space/options/proxy/ProxyReturnOptions.java
Elishtar: Why we can't? Looks like we can change method interface methods signature

src/main/java/io/tarantool/driver/core/proxy/CRUDBaseOptions.java
Elishtar: Okay, I agree

src/main/java/io/tarantool/driver/api/space/options/proxy/ProxyReturnOptions.java
ArtDu: Yeah we can change method signature, but in the inner implementation we need to do a hack to cast to `List<String>.class`

Elishtar (dismissed):

ArtDu:
We have one problem with our approach we can't have multiple inheritances in java and especially in our ProxyOptions.
For example, we have InsertOptions and CountOption, both have batch_id option, but only InsertOption has fields option
image

ArtDu:
Blocked by #272 because driver should parse metadata from crud. We can't get field by field name if we don't know position and It will frustrate users 
image

PR #267, opened 2022-09-17
Add Options parameter to Space operations

### LINKED ISSUE #266: Implement Options parameter to pass crud options

ArtDu:
Now we can add all options from crud to our Space API.

I didn't forget about:

- [x] Tests
- [x] Changelog
- [ ] Documentation

Closes #266

### DIFF
--- a/src/main/java/io/tarantool/driver/core/space/options/package-info.java
+++ b/src/main/java/io/tarantool/driver/core/space/options/package-info.java
@@ -1,0 +1,7 @@
+/**
+ * Contains implementations for Tarantool space options
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+package io.tarantool.driver.core.space.options;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Features
+- Added options parameter to Tarantool Space API ([#266](https://github.com/tarantool/cartridge-java/pull/266))
+
 ## [0.8.2] - 2022-09-16
 
 ### Features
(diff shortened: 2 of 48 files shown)


src/main/java/io/tarantool/driver/core/space/options/package-info.java
```diff
@@ -0,0 +1,7 @@
+/**
+ * Contains implementations for Tarantool space options
+ *
+ * @author Alexey Kuzin
+ * @author Artyom Dubinin
+ */
+package io.tarantool.driver.core.space.options;
```
akudiyar, line 7: We need to move the whole this package to `io.tarantool.driver.api.space.options` because these classes will be used by the customers and we will have to maintain the backward compatibility for them, as for all the other classes in the `api` package.

akudiyar (changes requested):
Cool, really like it, just two moments:
 - Let's move the Options classes (all package) to the `api` package. It is a part of the user-facing API.
 - Each `TarantoolSpaceOperations` method must have it's own derived `Options` class. So, for example, `delete` takes a `DeleteOptions` instance. Otherwise, our users will be able to put a wrong options instance into a method and become confused.

src/main/java/io/tarantool/driver/core/space/options/package-info.java
ArtDu: Moved

Elishtar:
Looks good, optionally maybe should add a little bit logging for operations, like "Client <client name> run operation <operation type>"

Elishtar (approved):

ArtDu:
> Looks good, optionally maybe should add a little bit logging for operations, like "Client run operation "

A log slows down request and I don't see an option why I need to know that client is going to tarantool. If client wanted it, it would add an additional log before the call operation

PR #274, opened 2022-09-29
Add example how to calculate bucketId on client

ArtDu:

I didn't forget about

- [x] Tests
- [ ] Changelog
- [x] Documentation

Closes #???

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,59 @@ class Scratch {
 }
 ```
 
+#### Using custom sharding function
+
+We can use your own sharding function to determine bucket id - location in cluster and use it in cluster operations.
+For this purpose you need:
+1) hash function  
+    For the example, equality of function from tarantool - [crc32](https://www.tarantool.io/en/doc/latest/reference/reference_lua/digest/#lua-function.digest.crc32) with specific polynomial value.
+    Java doesn't have crc32 out of the box with the ability to pass polynomial value, then we'll implement our own:
+    ```java
+    private static long crc32(byte[] data) {
+        BitSet bitSet = BitSet.valueOf(data);
+        int crc32 = 0xFFFFFFFF; // initial value
+        for (int i = 0; i < data.length * 8; i++) {
+            if (((crc32 >>> 31) & 1) != (bitSet.get(i) ? 1 : 0)) {
+                crc32 = (crc32 << 1) ^ 0x1EDC6F41; // xor with polynomial
+            } else {
+                crc32 = crc32 << 1;
+            }
+        }
+        crc32 = Integer.reverse(crc32); // result reflect
+        return crc32 & 0x00000000ffffffffL; // the unsigned java problem
+    }
+    ```
+2) number of buckets  
+   We can obtain number from tarantool via `vshard.router.bucket_count` function out of [vshard module](https://github.com/tarantool/vshard)
+    ```java
+   public static <T extends Packable, R extends Collection<T>> Integer getBucketCount(
+           TarantoolClient<T, R> client) throws ExecutionException, InterruptedException {
+       if (!bucketCount.isPresent()) {
+           bucketCount = Optional.ofNullable(
+                   client.callForSingleResult("vshard.router.bucket_count", Integer.class).get()
+           );
+       }
+       return bucketCount.get();
+   }
+    ```
+
+Then we can determine bucket id by passing your key through hash function and get the remainder of the division by number of buckets:
+```java
+byte[] key = ... // can be multipart
+Integer bucketId = (bucketIdcrc32(key) % getBucketCount(client)) + 1;
+```
+
+After that we may apply it in operations:
+```java
+InsertOptions insertOptions = ProxyInsertOptions.create().withBucketId(bucketId);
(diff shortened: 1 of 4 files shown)


ArtDu:
@akudiyar @Elishtar 

@dkasimovskiy suggests to export this implementation in source. What do you think about it?

akudiyar:
I think a good solution will be to describe how to create and use your own sharding function with the help of vshard API calls in the README. And leave the sample code in tests how they are now.

src/test/java/io/tarantool/driver/integration/Utils.java
```diff
@@ -1,30 +1,97 @@
...
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+        return Math.toIntExact(
+                (crc32(outputStream.toByteArray()) % getBucketCount(client)) + 1
+        );
+    }
+
+    private static long crc32(byte[] data) {
```
akudiyar, line 84: If this function was borrowed from any public place, it is a good manner to reference the source.

ArtDu: Added

ArtDu:
> I think a good solution will be to describe how to create and use your own sharding function with the help of vshard API calls in the README. And leave the sample code in the tests how it is now.

Added, please validate it

README.md
```diff
@@ -244,6 +244,59 @@ class Scratch {
...
+   We can obtain number from tarantool via `vshard.router.bucket_count` function out of [vshard module](https://github.com/tarantool/vshard)
+    ```java
+   public static <T extends Packable, R extends Collection<T>> Integer getBucketCount(
+           TarantoolClient<T, R> client) throws ExecutionException, InterruptedException {
+       if (!bucketCount.isPresent()) {
+           bucketCount = Optional.ofNullable(
+                   client.callForSingleResult("vshard.router.bucket_count", Integer.class).get()
+           );
+       }
+       return bucketCount.get();
```
akudiyar, line 279: I think we need to handle an unexpected `null` value better, like with an `orElseThrow` callback

README.md
```diff
@@ -244,6 +244,59 @@ class Scratch {
...
+   }
+    ```
+
+Then we can determine bucket id by passing your key through hash function and get the remainder of the division by number of buckets:
+```java
+byte[] key = ... // can be multipart
```
akudiyar, line 285: It's not clear now what to put in the key. An example with a tuple will be better

README.md
```diff
@@ -244,6 +244,59 @@ class Scratch {
...
+After that we may apply it in operations:
+```java
+InsertOptions insertOptions = ProxyInsertOptions.create().withBucketId(bucketId);
+insertResult = profileSpace.insert(tarantoolTuple, insertOptions).get();
+
+ProxySelectOptions selectOptions = ProxySelectOptions.create().withBucketId(bucketId);
+selectResult = profileSpace.select(condition, selectOptions).get();
+```
+
+You can see sources of example in [tests](src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceInsertOptionsIT.java)
```
akudiyar, line 298:
```suggestion
You can see the sources of this example in the [tests](src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceInsertOptionsIT.java)
```

akudiyar (changes requested):
LGTM,just two points here:
 - an example of converting a tuple to bytes will be useful
 - fixed the grammar a bit

README.md
ArtDu: Rewritten, please check it out

README.md
ArtDu: Rewritten, please check it out

akudiyar (approved):
LGTM

README.md
```diff
@@ -244,6 +244,60 @@ class Scratch {
...
+   This number can be obtained from Tarantool via `vshard.router.bucket_count` function out of [vshard module](https://github.com/tarantool/vshard)
+    ```java
+   public static <T extends Packable, R extends Collection<T>> Integer getBucketCount(
+           TarantoolClient<T, R> client) throws ExecutionException, InterruptedException {
+       if (!bucketCount.isPresent()) {
+           bucketCount = Optional.ofNullable(
+                   client.callForSingleResult("vshard.router.bucket_count", Integer.class).get()
+           );
+       }
+       bucketCount.orElseThrow(() -> new TarantoolClientException("Failed to get bucket count"));
```
akudiyar, line 279: `return` is missing

README.md
```diff
@@ -244,6 +244,60 @@ class Scratch {
...
+   }
+    ```
+
+Then we can determine bucket id by passing your key through hash function and get the remainder of the division by number of buckets:
+```java
+TarantoolTuple tarantoolTuple = tupleFactory.create(1, null, "FIO", 50, 100);
+byte[] key = getBytesFromList(Arrays.asList(tarantoolTuple.getInteger(0), tarantoolTuple.getInteger(2)));
```
akudiyar, line 286: It may be good to place a link to this `getBytesFromList` method in the tests to the comment above, because it is not a built-in and therefore produces questions

akudiyar (changes requested):
Two small changes are needed

PR #275, opened 2022-10-02
Changes for TarantoolNullField class to singleton

### LINKED ISSUE #195: Make TarantoolNullField class static
Currently we create a new object every time we have a **null** field in a tuple.
Due to performance improvement we need to make behaviour of **TarantoolNullField** class similar to `Optional.empty()`.
Just to have one TarantoolNullField object for all application.

This is a BRAKING CHANGE!

devrishal:
The `TarantoolNullField `class was not singleton and its only task it to provide functionality similar to `Optional.Empty()`.
Changes done to make `TarantoolNullField `behaves same as `Optional.Empty()`.

I didn't forget about

- [X] Tests
- [X] Changelog
- [X] Documentation
- [X] Cleanup the code for review. See [checklist](https://github.com/tarantool/cartridge-java/blob/master/docs/review-checklist.md)

Closes #195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased]
-
+- Changed TarantoolNullField class to singleton ([#195](https://github.com/tarantool/cartridge-java/pull/275))
 ## [0.9.0] - 2022-10-03
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 - Changed TarantoolNullField class to singleton ([#195](https://github.com/tarantool/cartridge-java/pull/275))
+
 ## [0.9.0] - 2022-10-03
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased]
-
+- Changed TarantoolNullField class to singleton
 ### Features
 - Added options parameter to Tarantool Space API ([#266](https://github.com/tarantool/cartridge-java/pull/266))
 - Added bucket id parameter to Tarantool Space API ([#270](https://github.com/tarantool/cartridge-java/pull/270))
--- a/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
+++ b/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
@@ -12,7 +12,7 @@ import org.msgpack.value.ValueFactory;
  */
 public final class TarantoolNullField implements TarantoolField {
 
-    public static final TarantoolNullField EMPTY = new TarantoolNullField();
+    public static final TarantoolNullField INSTANCE = new TarantoolNullField();
 
     private TarantoolNullField() {
     }
--- a/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
+++ b/src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
@@ -12,15 +12,9 @@ import org.msgpack.value.ValueFactory;
  */
 public final class TarantoolNullField implements TarantoolField {
 
-    private static final TarantoolNullField EMPTY = new TarantoolNullField(null);
-    private final TarantoolNullField value;
+    public static final TarantoolNullField EMPTY = new TarantoolNullField();
 
-    private TarantoolNullField(TarantoolNullField value) {
-        this.value = value;
-    }
-
-    public static TarantoolNullField create() {
-        return EMPTY;
+    private TarantoolNullField() {
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


akudiyar (changes requested):
Hi! Thanks for the patch!

I have a couple of questions:

1. I suggest using `create()` as the factory method name. With the `Optional` class we can create instances containing either null or non-null values, but with this class, there is only one possible internal value. So the name `empty()` can imply that something "non-empty" exists too, while `create()` can not.
2. Please fix the incorrect whitespace and run `mvn verify` to fix the other style warnings. 

UPD: discussed the change internally with team, I see that this change may be useful.

devrishal:
@akudiyar Modified the code as per the review comments and also changed for checkstyle.

ArtDu (approved):
LGTM

ArtDu:
@devrisha I forgot about changelog, can you write one line about changes in CHANGELOG.md?

devrishal:
> @devrisha I forgot about changelog, can you write one line about changes in CHANGELOG.md?

Yes will do now.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased]
-
+- Changed TarantoolNullField class to singleton
```
ArtDu, line 4: But you must rebase, because we updated changelog today

devrishal: Please check , did a rebase with the latest changes.

dkasimovskiy (changes requested):

src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
```diff
@@ -12,6 +12,17 @@
  */
 public final class TarantoolNullField implements TarantoolField {
 
+    private static final TarantoolNullField EMPTY = new TarantoolNullField(null);
```
Elishtar, line 15: I suppose we need chose one from two strategy
Not lazy initialization
`public static final TarantoolNullField EMPTY = new TarantoolNullField(null);`
or
lazy initialization
```
private static TarantoolNullField empty;

public static TarantoolNullField create() {
       if (empty == null) 
        empty = new TarantoolNullField(null);;
    }
    
   return empty
}
```

devrishal: @Elishtar Its already` static final ` and will be immediately instance is assigned to it. The `TarantoolNullField ` is already a final and immutable class. Not sure if you are considering the Thread safety or in any other sense? But to answer the Thread safety the above points can be considered.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [Unreleased]
-
+- Changed TarantoolNullField class to singleton ([#195](https://github.com/tarantool/cartridge-java/pull/275))
```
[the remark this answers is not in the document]
devrishal: Added new line as per suggestion.

src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
Elishtar: @devrishal if you already initialize  EMPTY, not necessarily use method create for access just change privat to public for EMPTY, if you want use method create. for access you can initialize EMPTY at the moment of first access please chose one strategy, now you use combination

dkasimovskiy (approved):

src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
devrishal: Please let me know if this is fine for you ```
private static final TarantoolNullField EMPTY = new TarantoolNullField();
    private TarantoolNullField() {
    }

    public static TarantoolNullField create() {
        return EMPTY;
    }
```

akudiyar (approved):
LGTM.

I agree that lazy init is not needed here, it gives no advantages.

akudiyar:
@devrishal apologies for making that many comments for a simple change, but what do you think about dropping the `create()` method at all and leaving just the public constant? It looks like the method is redundant here since it always returns the constant.
Also, a classic name for such a singleton constant is usually `INSTANCE`.

src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
devrishal: @akudiyar @Elishtar Please check and let me know if the changes looks good.

devrishal:
> @devrishal apologies for making that many comments for a simple change, but what do you think about dropping the `create()` method at all and leaving just the public constant? It looks like the method is redundant here since it always returns the constant. Also, a classic name for such a singleton constant is usually `INSTANCE`.

No Issues, It is indeed a good learning for me :) . Please let me know if the changes are fine now.

src/main/java/io/tarantool/driver/api/tuple/TarantoolNullField.java
devrishal: Please check if the changes looks good?

Elishtar (approved):

PR #278, opened 2022-10-03
Add parsing metadata from crud response

### LINKED ISSUE #272: Parse metadata from crud response
image

ArtDu:
Now we can obtain metadata from the response and don't use metadata from ddl.

I didn't forget about:

- [x] Tests
- [x] Changelog
- [ ] Documentation
- [x] Clean up the code for review. See [checklist](https://github.com/tarantool/cartridge-java/blob/master/docs/review-checklist.md)

Closes #272


src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
```diff
@@ -57,16 +66,68 @@ public TarantoolResultImpl(Value value, ValueConverter<ArrayValue, T> tupleConve
         }
     }
 
+    private TarantoolSpaceMetadata parseCRUDMetadata(Value metadata) {
+        if (!metadata.isArrayValue()) {
```
Elishtar, line 70: Looks like potential NullPointerException, I guess will be good idea add fast fail if metadata is null

src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
```diff
@@ -57,16 +66,68 @@ public TarantoolResultImpl(Value value, ValueConverter<ArrayValue, T> tupleConve
...
+        for (Value value : responseFormat) {
+            if (!value.isMapValue()) {
+                throw new TarantoolClientException("Unsupported crud metadata format: expected array of maps");
+            }
+            Map<Value, Value> fieldMap = value.asMapValue().map();
+
+            Value fieldNameValue = fieldMap.get(FORMAT_NAME_KEY);
+            if (fieldNameValue == null || !fieldNameValue.isStringValue()) {
+                throw new TarantoolClientException(
+                        "Unsupported space metadata format: key '" + FORMAT_NAME_KEY + "' must have string value");
```
Elishtar, line 87: add current value to exception log

src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
```diff
@@ -57,16 +66,68 @@ public TarantoolResultImpl(Value value, ValueConverter<ArrayValue, T> tupleConve
...
+            }
+            String fieldName = fieldNameValue.asStringValue().asString();
+
+            Value fieldTypeValue = fieldMap.get(FORMAT_TYPE_KEY);
+            if (fieldTypeValue == null || !fieldTypeValue.isStringValue()) {
+                throw new TarantoolClientException(
+                        "Unsupported space metadata format: key '" + FORMAT_TYPE_KEY + "' must have string value");
```
Elishtar, line 94: The same

ArtDu:
The main problem is that now we have two converters used to parse incoming tuples:
The first level parses the top level - container - List of tuples or crud response.
The second level parses one tuple at a time.

The second level always needs to know the metadata, because it creates a TarantoolTuple and passes it there. Right now it only gets the metadata from the constructor.
The conclusion is the level that parses a particular tuple must know the metadata from the level above, if it is crudResponse.

So I see 3 options:
1) Create a new fromValue method, into which we can pass custom metadata, as I did in PR - minus that we add a method to the API, although the method is not giant terrible)
2) Combine the first and second levels - consider combining two converters into one - minus, that everything is in one place parsing the container and the tuples themselves
3) Create a converter for each request of the second level - minus the memory will be painfull grown

src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
ArtDu: Fixed

akudiyar:
I think the new converter variant (3) is the because it doesn't allow to leak the implementation details through interfaces and leaves in place the ability to have different result mappers (e.g. for libs other than Tarantool/crud or non-tuple results).

The converter instantiation is a parallel problem and should be dealt with as a whole, not in this particular place.

src/main/java/io/tarantool/driver/mappers/MultiValueTarantoolResultMapperFactory.java
```diff
@@ -82,4 +84,11 @@ public MultiValueTarantoolResultMapperFactory(MessagePackMapper messagePackMappe
                                  Class<? extends MultiValueCallResult<T, TarantoolResult<T>>> resultClass) {
         return withMultiValueResultConverter(new TarantoolResultConverter<>(valueConverter), resultClass);
     }
+
+    public CallResultMapper<TarantoolResult<T>, MultiValueCallResult<T, TarantoolResult<T>>>
+    withTarantoolTupleResultConverter(TarantoolTupleConverter valueConverter,
+                                      Class<? extends MultiValueCallResult<T, TarantoolResult<T>>> resultClass) {
+        return withMultiValueResultConverter((ValueConverter) new TarantoolTupleResultConverter<>(valueConverter),
```
akudiyar, line 91: Don't understand why a cast is needed here. Probably the class needs more proper generics definition or something.

src/main/java/io/tarantool/driver/mappers/TarantoolTupleMultiResultMapperFactory.java
```diff
@@ -42,7 +42,7 @@ public TarantoolTupleMultiResultMapperFactory(MessagePackMapper messagePackMappe
     public CallResultMapper<TarantoolResult<TarantoolTuple>,
             MultiValueCallResult<TarantoolTuple, TarantoolResult<TarantoolTuple>>>
     withDefaultTupleValueConverter(MessagePackMapper messagePackMapper, TarantoolSpaceMetadata spaceMetadata) {
-        return withTarantoolResultConverter(
+        return withTarantoolTupleResultConverter(
```
akudiyar, line 45: Isn't this change breaking the plain box client?

src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
```diff
@@ -57,19 +68,19 @@ public TarantoolResultImpl(Value value, ValueConverter<ArrayValue, T> tupleConve
...
         this.tuples = tupleArray.list().stream()
-            .map(v -> {
-                try {
-                    return tupleConverter.fromValue(v.asArrayValue());
-                } catch (MessageTypeCastException e) {
-                    throw new TarantoolTupleConversionException(v, e);
-                }
-            })
-            .collect(Collectors.toList());
+                .map(v -> {
```
akudiyar, line 73: nit: Please set up 4 spaces as continuation identation in Idea/whatever. These unrelated space changes distract a lot.

akudiyar (changes requested):
It's a better idea overall then changing the ValueConverter interface, however the changes look very rough. I'd rather not copy-paste the whole converters code in that way. Also It may be worth thinking about passing a tuple converter factory to the result converter instead of a tuple converter, that will help eliminating the fromValue method and unnecessary casts.

src/main/java/io/tarantool/driver/core/TarantoolResultImpl.java
```diff
@@ -57,19 +68,19 @@ public TarantoolResultImpl(Value value, ValueConverter<ArrayValue, T> tupleConve
...
+                    try {
```
Elishtar, line 74: I guess will be better extract try/catch to method

src/main/java/io/tarantool/driver/core/TarantoolTupleResultImpl.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Basic TarantoolResult implementation. Supports two types of the tuple result: a MessagePack array of arrays
+ * (as in IPROTO) and a map with format {@code {"rows": [[], ...], "metadata": []}} as returned by tarantool/crud module
+ *
+ * @author Alexey Kuzin
+ */
+public class TarantoolTupleResultImpl extends TarantoolResultImpl<TarantoolTuple> {
+
+    public TarantoolTupleResultImpl(Value value, TarantoolTupleConverter tupleConverter) {
+        if (value.isArrayValue()) {
```
Elishtar, line 30: Let's try to replace multiple if/else to case for example like this
https://stackoverflow.com/questions/5579309/is-it-possible-to-use-the-instanceof-operator-in-a-switch-statement

src/main/java/io/tarantool/driver/core/TarantoolTupleResultImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        if (metadata == null || !metadata.isArrayValue()) {
+            throw new TarantoolClientException("Unsupported crud metadata format: expected array of maps");
+        }
+
+        List<Value> responseFormat = metadata.asArrayValue().list();
+
+        Map<String, TarantoolFieldMetadata> spaceFormatMetadata = new LinkedHashMap<>();
+        int fieldPosition = 0;
+        for (Value value : responseFormat) {
+            if (!value.isMapValue()) {
```
Elishtar, line 65: Case maybe?

src/main/java/io/tarantool/driver/core/TarantoolTupleResultImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        spaceMetadata.setSpaceFormatMetadata(spaceFormatMetadata);
+
+        return spaceMetadata;
+    }
+
+    private void setTuples(ArrayValue tupleArray, TarantoolTupleConverter tupleConverter,
+                           TarantoolSpaceMetadata responseMetadata) {
+        this.tuples = tupleArray.list().stream()
+                .map(v -> {
+                    try {
```
Elishtar, line 100: Maybe extract

src/main/java/io/tarantool/driver/mappers/converters/value/custom/TarantoolTupleResultConverter.java
```diff
@@ -0,0 +1,29 @@
...
+import org.msgpack.value.ArrayValue;
+import org.msgpack.value.Value;
+
+/**
+ * @author Alexey Kuzin
+ */
+public class TarantoolTupleResultConverter<V extends Value>
+        implements ValueConverter<V, TarantoolResult<TarantoolTuple>> {
+
+    private static final long serialVersionUID = 20200708L;
```
Elishtar, line 17: Auto generate?

src/main/java/io/tarantool/driver/core/metadata/parsers/CRUDResponseMetadataParser.java
```diff
@@ -0,0 +1,16 @@
+package io.tarantool.driver.core.metadata.parsers;
+
+import io.tarantool.driver.core.metadata.TarantoolSpaceMetadataImpl;
+import org.msgpack.value.Value;
+
+/**
+ * @author Artyom Dubinin
+ */
+public class CRUDResponseMetadataParser implements SpaceMetadataParser {
```
akudiyar, line 9: What's the purpose of this class? Currently it looks like doing nothing.

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
```diff
@@ -0,0 +1,40 @@
...
+import io.tarantool.driver.core.metadata.TarantoolSpaceMetadataImpl;
+import io.tarantool.driver.exceptions.TarantoolClientException;
+import org.msgpack.value.Value;
+
+import java.util.Map;
+
+/**
+ * @author Artyom Dubinin
+ */
+public class DDLSpaceMetadataParser implements SpaceMetadataParser {
```
akudiyar, line 13: This class needs some unit tests and a Javadoc.

src/main/java/io/tarantool/driver/core/metadata/parsers/SpaceMetadataParser.java
```diff
@@ -0,0 +1,52 @@
...
+import org.msgpack.value.impl.ImmutableStringValueImpl;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * @author Artyom Dubinin
+ */
+public interface SpaceMetadataParser {
```
akudiyar, line 19: This class needs some unit tests and a javadoc.

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public TarantoolSpaceMetadataImpl parse(Value metadata) {
```
akudiyar, line 16: Shouldn't it take `MapValue` explicitly?

src/main/java/io/tarantool/driver/core/metadata/parsers/VSpaceMetadataParser.java
```diff
@@ -0,0 +1,43 @@
...
+import io.tarantool.driver.core.metadata.TarantoolSpaceMetadataImpl;
+import io.tarantool.driver.mappers.MessagePackValueMapper;
+import org.msgpack.value.Value;
+
+import java.util.Iterator;
+
+/**
+ * @author Artyom Dubinin
+ */
+public class VSpaceMetadataParser implements SpaceMetadataParser {
```
akudiyar, line 12: This class needs unit tests and a Javadoc.

akudiyar (changes requested):

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
ArtDu: If we want to have one interface for metadata parsing we need to use `Value`. Metadata can be stored in different structures

src/main/java/io/tarantool/driver/core/metadata/parsers/CRUDResponseMetadataParser.java
ArtDu: I changed classes. You should check

src/main/java/io/tarantool/driver/core/metadata/parsers/SpaceMetadataParser.java
ArtDu: Fixed

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
ArtDu: I changed it

src/main/java/io/tarantool/driver/core/metadata/parsers/VSpaceMetadataParser.java
ArtDu: Fixed

ArtDu:
We can split PR to "Parse metadata from crud response" 3252ba6cd60b0e09661c378263eeac65f7799109 and "Refactor metadata parsing". But I prefer to merge "Parse metadata from crud response" 3252ba6cd60b0e09661c378263eeac65f7799109 for first

src/main/java/io/tarantool/driver/core/metadata/TarantoolFieldMetadataImpl.java
```diff
@@ -8,7 +8,7 @@
  * @author Sergey Volgin
  * @author Artyom Dubinin
  */
-class TarantoolFieldMetadataImpl implements TarantoolFieldMetadata {
+public class TarantoolFieldMetadataImpl implements TarantoolFieldMetadata {
```
akudiyar, line 11: Do we really need to make it public? It's an internal implementation class.

akudiyar (changes requested):
Interestingly, that looks pretty OK without all other changes, one question is left though

src/main/java/io/tarantool/driver/core/metadata/TarantoolFieldMetadataImpl.java
ArtDu: There are two options here:
* We store these converters(parsers in future) in `*.core.metadata` package
* Or we open this class and store classes where they should be semantically

ArtDu: First option has been picked up

akudiyar (approved):
LGTM

PR #290, opened 2022-10-22
long array mapper

anatennis:
WorkInProgress

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [x] Documentation

Related issues:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Adding default mapper for long arrays ([#290](https://github.com/tarantool/cartridge-java/pull/290))
+
 ## [0.9.1] - 2022-10-13
 
 - Changed TarantoolNullField class to singleton ([#195](https://github.com/tarantool/cartridge-java/pull/275))
--- a/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapperFactory.java
+++ b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapperFactory.java
@@ -1,8 +1,45 @@
 package io.tarantool.driver.mappers;
 
-import io.tarantool.driver.mappers.converters.object.*;
-import io.tarantool.driver.mappers.converters.value.*;
-import org.msgpack.value.*;
+import io.tarantool.driver.mappers.converters.object.DefaultNilValueToNullConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultBigDecimalToExtensionValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultLongArrayToArrayValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultBooleanToBooleanValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultByteArrayToBinaryValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultCharacterToStringValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultDoubleToFloatValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultFloatToFloatValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultIntegerToIntegerValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultLongToIntegerValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultPackableObjectConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultShortToIntegerValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultStringToStringValueConverter;
+import io.tarantool.driver.mappers.converters.object.DefaultUUIDToExtensionValueConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultArrayValueToLongArrayConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultBinaryValueToByteArrayConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultBooleanValueToBooleanConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultExtensionValueToBigDecimalConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultExtensionValueToUUIDConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultFloatValueToDoubleConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultFloatValueToFloatConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultFloatValueToIntegerConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultFloatValueToLongConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultFloatValueToShortConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultIntegerValueToIntegerConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultIntegerValueToDoubleConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultIntegerValueToFloatConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultIntegerValueToLongConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultIntegerValueToShortConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultStringValueToByteArrayConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultStringValueToCharacterConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultStringValueToStringConverter;
+import org.msgpack.value.ArrayValue;
+import org.msgpack.value.BinaryValue;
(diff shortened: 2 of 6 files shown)


src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapperFactory.java
```diff
@@ -78,13 +44,15 @@ private DefaultMessagePackMapperFactory() {
                 .withValueConverter(ValueType.EXTENSION, BigDecimal.class,
                         new DefaultExtensionValueToBigDecimalConverter())
                 .withValueConverter(ValueType.NIL, Object.class, new DefaultNilValueToNullConverter())
-                //TODO: Potential issue https://github.com/tarantool/cartridge-java/issues/118
+                //TODO: add this when will it be resolved https://github.com/tarantool/cartridge-java/issues/118
```
akudiyar, line 47: Please rebase on the fresh master branch or remove these changes, this has been already changed.

src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapperFactory.java
```diff
@@ -1,42 +1,8 @@
...
-import org.msgpack.value.BinaryValue;
-import org.msgpack.value.BooleanValue;
-import org.msgpack.value.ExtensionValue;
-import org.msgpack.value.FloatValue;
-import org.msgpack.value.IntegerValue;
-import org.msgpack.value.StringValue;
-import org.msgpack.value.ValueType;
+import io.tarantool.driver.mappers.converters.object.*;
+import io.tarantool.driver.mappers.converters.value.*;
+import org.msgpack.value.*;
```
akudiyar, line 5: Please do not use the "star" import. In Intellij IDEA you can tick this option in the code autoformat settings.

src/main/java/io/tarantool/driver/mappers/converters/object/DefaultLongArrayToArrayValueConverter.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Arrays;
+
+/**
+ * Default {@code long[]} to {@link ArrayValue} converter
+ */
+public class DefaultLongArrayToArrayValueConverter implements ObjectConverter<long[], ArrayValue> {
+
+    @Override
+    public ArrayValue toValue(long[] object) {
+        return ValueFactory.newArray(toNumberValue(object), true);
```
akudiyar, line 18: I think `toNumberValue` will not work like this. Since it is a method for a single value, not array.

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultArrayValueToLongArrayConverter.java
```diff
@@ -0,0 +1,27 @@
...
+        long[] values = new long[value.size()];
+        for (int i = 0; i < value.size(); i++) {
+            values[i] = value.list().get(i).asNumberValue().toLong();
+        }
+        return values;
+    }
+
+    @Override
+    public boolean canConvertValue(ArrayValue value) {
+        return value.list().stream().allMatch(Value::isNumberValue);
```
akudiyar, line 24: q: I believe `toLongValue` can be better here... But for saving time we may check only the first value (assuming that we have only the "correct" arrays on the Tarantool side). @ArtDu @Elishtar do you have any arguments for checking all array elements here?

akudiyar (changes requested):
Hi, thanks for the patch! Please add test and addresss the comments below.

src/main/java/io/tarantool/driver/mappers/converters/object/DefaultLongArrayToArrayValueConverter.java
anatennis: It's my method, so I've changed its name to toNumberValueArray

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultArrayValueToLongArrayConverter.java
anatennis: Actually any number can be converted to long (int, short, double) - so my thoughts were like `asNumberValue` does the same check and it will suit for this case. Or not?
And also I have doubts about only one element checking. Because this array is also valid `arrayValue = ValueFactory.newArray(new ImmutableLongValueImpl(2L), new ImmutableStringValueImpl("string"), new ImmutableLongValueImpl(3L));` and if we'll check only 1st element - then `canConvertValue()` returns true but later `fromValue()` will throw `MessageTypeCastException`

src/test/java/io/tarantool/driver/mappers/DefaultLongArrayConverterTest.java
```diff
@@ -0,0 +1,47 @@
...
+import io.tarantool.driver.mappers.converters.object.DefaultLongArrayToArrayValueConverter;
+import io.tarantool.driver.mappers.converters.value.DefaultArrayValueToLongArrayConverter;
+import org.junit.jupiter.api.Test;
+import org.msgpack.value.ArrayValue;
+import org.msgpack.value.ImmutableArrayValue;
+import org.msgpack.value.ValueFactory;
+import org.msgpack.value.impl.ImmutableLongValueImpl;
+import org.msgpack.value.impl.ImmutableStringValueImpl;
+
+import static org.junit.jupiter.api.Assertions.*;
```
akudiyar, line 12: Please fix the star import

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultArrayValueToLongArrayConverter.java
```diff
@@ -0,0 +1,30 @@
...
+        for (int i = 0; i < value.size(); i++) {
+            values[i] = value.list().get(i).asNumberValue().toLong();
+        }
+        return values;
+    }
+
+    @Override
+    public boolean canConvertValue(ArrayValue value) {
+        if (value != null) {
+            return value.list().stream().allMatch(Value::isNumberValue);
```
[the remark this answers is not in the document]
akudiyar: I suggest adding a test case for wrong value as well

akudiyar (changes requested):
Looks better now, please add some test cases to the https://github.com/tarantool/cartridge-java/blob/master/src/test/java/io/tarantool/driver/mappers/DefaultMessagePackMapperTest.java test suite too. We need to check the correctness of convertation if a result tuple or a map contains different `ArrayValue` elements.

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultArrayValueToLongArrayConverter.java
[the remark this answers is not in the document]
anatennis: Ok, no problem, now it checks only 1st value. Test was already added but now it checks that `canConvertValue` returns true and `fromValue` throws Exception

anatennis:
Added also tests to DefaultMessagePackMapperTest

src/main/java/io/tarantool/driver/mappers/converters/object/DefaultLongArrayToArrayValueConverter.java
```diff
@@ -0,0 +1,26 @@
...
+import org.msgpack.value.impl.ImmutableLongValueImpl;
+
+import java.util.Arrays;
+
+/**
+ * Default {@code long[]} to {@link ArrayValue} converter
+ *
+ * @author Anastasiia Romanova
+ */
+public class DefaultLongArrayToArrayValueConverter implements ObjectConverter<long[], ArrayValue> {
```
akudiyar, line 16: Looks like the `serialVersionUID` variable is missing here.

Please set its value to the current date.

src/main/java/io/tarantool/driver/mappers/converters/value/DefaultArrayValueToLongArrayConverter.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.tarantool.driver.mappers.converters.ValueConverter;
+import org.msgpack.value.ArrayValue;
+
+/**
+ * Default {@link ArrayValue} to {@code long[]} converter
+ *
+ * @author Anastasiia Romanova
+ */
+public class DefaultArrayValueToLongArrayConverter implements ValueConverter<ArrayValue, long[]> {
```
akudiyar, line 11: Please add the `serialVersionUID` value here too

akudiyar (changes requested):
Please run `mvn verify` and fix the linter warnings.

akudiyar (commented):
Don't forget to add this change to CHANGELOG

anatennis:
fix 

> Please run `mvn verify` and fix the linter warnings.

done

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+## [0.9.1] - 2022-10-26
```
akudiyar, line 5: nit: an extra line

akudiyar (approved):
LGTM, let's fix that one line in CHANGELOG if you have time.

akudiyar (approved):
LGTM

PR #322, opened 2022-12-12
Add TarantoolTupleResultMapper factory

### LINKED ISSUE #321: Add simple to use TarantoolTupleMapper factory
Cause it's so hard to use current factories

ref: https://github.com/tarantool/cartridge-java/pull/300#issuecomment-1328692045

ArtDu:
We need it to create universal mapper that can parse crud and box responses.
Also with this factory we can't see inherited methods.

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes #321

### DIFF
--- a/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
+++ b/src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
@@ -67,6 +67,7 @@ public abstract class AbstractTarantoolClient<T extends Packable, R extends Coll
     private final TarantoolConnectionListeners listeners;
     private final AtomicReference<TarantoolMetadata> metadataHolder = new AtomicReference<>();
     private final ResultMapperFactoryFactoryImpl mapperFactoryFactory;
+
     private final SpacesMetadataProvider metadataProvider;
     private final ScheduledExecutorService timeoutScheduler;
     private TarantoolConnectionManager connectionManager;
--- a/src/main/java/io/tarantool/driver/mappers/TarantoolTupleResultMapperFactory.java
+++ b/src/main/java/io/tarantool/driver/mappers/TarantoolTupleResultMapperFactory.java
@@ -1,0 +1,141 @@
+package io.tarantool.driver.mappers;
+
+import io.tarantool.driver.api.MultiValueCallResult;
+import io.tarantool.driver.api.SingleValueCallResult;
+import io.tarantool.driver.api.TarantoolResult;
+import io.tarantool.driver.api.metadata.TarantoolSpaceMetadata;
+import io.tarantool.driver.api.tuple.TarantoolTuple;
+
+/**
+ * Provides factory that creates result mappers for parsing {@code TarantoolResult<TarantoolTuple>} from various
+ * structures
+ *
+ * @author Artyom Dubinin
+ */
+public interface TarantoolTupleResultMapperFactory {
+    /**
+     * Mapper parses result from a list of tuples.
+     * Use this mapper for handling containers with {@link TarantoolTuple} inside.
+     * The IProto method results by default contain lists of tuples.
+     * This mapper can be used at the inner level by other mapper factories that are handling higher-level containers.
+     * <p>
+     * input: array of tuples with MessagePack values inside ([t1, t2, ...])
+     * <br>
+     * mapper result: {@code TarantoolResult<TarantoolTuple>}
+     * <p>
+     * Returned TarantoolResult doesn't store metadata, and you can't get field by field name
+     *
+     * @return mapper that parses list of arrays to {@code TarantoolResult<TarantoolTuple>}
+     */
+    TarantoolResultMapper<TarantoolTuple> withArrayValueToTarantoolTupleResultConverter(
+        MessagePackMapper messagePackMapper);
+
+    /**
+     * Mapper parses result from a list of tuples.
+     * Use this mapper for handling containers with {@link TarantoolTuple} inside.
+     * The IProto method results by default contain lists of tuples.
+     * This mapper can be used at the inner level by other mapper factories that are handling higher-level containers.
+     * <p>
+     * input: array of tuples with MessagePack values inside ([t1, t2, ...])
+     * <br>
+     * mapper result: {@code TarantoolResult<TarantoolTuple>}
+     * <p>
+     * For example, it can be used to parse result from IPROTO_SELECT
+     *
+     * @return mapper that parses list of arrays to {@code TarantoolResult<TarantoolTuple>}
+     */
+    TarantoolResultMapper<TarantoolTuple> withArrayValueToTarantoolTupleResultConverter(
+        MessagePackMapper messagePackMapper, TarantoolSpaceMetadata spaceMetadata);
+
+    /**
+     * Mapper parses result from a map with list of tuples and metadata.
+     * Use this mapper for handling containers with {@link TarantoolTuple} inside.
+     * The crud method results by default contain map with list of tuples and metadata.
+     * This mapper can be used at the inner level by other mapper factories that are handling higher-level containers.
+     * <p>
+     * input: map with metadata and array of tuples with MessagePack values inside ([t1, t2, ...])
+     * <br>
+     * mapper result: {@code TarantoolResult<TarantoolTuple>}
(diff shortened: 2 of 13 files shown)


src/main/java/io/tarantool/driver/mappers/TarantoolTupleResultMapperFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import io.tarantool.driver.api.MultiValueCallResult;
+import io.tarantool.driver.api.SingleValueCallResult;
+import io.tarantool.driver.api.TarantoolResult;
+import io.tarantool.driver.api.metadata.TarantoolSpaceMetadata;
+import io.tarantool.driver.api.tuple.TarantoolTuple;
+
+/**
+ * @author Artyom Dubinin
+ */
+public interface TarantoolTupleResultMapperFactory {
```
akudiyar, line 12: Javadocs are missing. The docs are needed both for the class (with example usage) and for the individual methods (with a purpose description)

src/main/java/io/tarantool/driver/mappers/TarantoolTupleResultMapperFactoryImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import io.tarantool.driver.mappers.factories.ArrayValueToTarantoolTupleResultMapperFactory;
+import io.tarantool.driver.mappers.factories.MultiValueWithTarantoolTupleResultMapperFactory;
+import io.tarantool.driver.mappers.factories.ResultMapperFactoryFactoryImpl;
+import io.tarantool.driver.mappers.factories.RowsMetadataToTarantoolTupleResultMapperFactory;
+import io.tarantool.driver.mappers.factories.SingleValueWithTarantoolTupleResultMapperFactory;
+
+/**
+ * @author Artyom Dubinin
+ */
+public class TarantoolTupleResultMapperFactoryImpl implements TarantoolTupleResultMapperFactory {
```
akudiyar, line 17: This class can implement a Singleton pattern (like the DefaultMessagePackMapperFactory)

src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
```diff
@@ -48,11 +49,12 @@ private void initClient() {
             .build();
 
         ResultMapperFactoryFactoryImpl factory = new ResultMapperFactoryFactoryImpl();
```
akudiyar, line 51: This variable doesn't seem to be used further

akudiyar (changes requested):
Looks OK, but Javadocs are missing

src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
ArtDu: Fixed

src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
```diff
@@ -111,6 +115,7 @@ public AbstractTarantoolClient(TarantoolClientConfig config, TarantoolConnection
 
         this.config = config;
         this.mapperFactoryFactory = new ResultMapperFactoryFactoryImpl();
+        this.resultMapperFactory = TarantoolTupleResultMapperFactoryImpl.getInstance();
```
akudiyar, line 118: It does not look right to put a concrete tuple factory implementation in the abstract client, given that the Tarantool tuple class is a generic parameter. I believe it should be instantiated in the corresponding final class or some intermediate one. In the root abstract class, we may put an abstract method for it with a generic type.

src/main/java/io/tarantool/driver/api/TarantoolCallOperations.java
```diff
@@ -491,4 +492,6 @@ <T, R extends List<T>> CompletableFuture<R> callForMultiResult(
      * @return result mapper factory instances factory instance
      */
     ResultMapperFactoryFactory getResultMapperFactoryFactory();
+
+    TarantoolTupleResultMapperFactory getTarantoolTupleResultMapperFactory();
```
akudiyar, line 496: Javadoc is missing

akudiyar (changes requested):
One important test kind that is missing - we need to test that several consequent and concurrent invocations of `TarantoolTupleResultMapperFactory` methods do not cause any problems with wrong mapper stacks. That means (until we haven't implemented any smart caching) a new isolated mapping stack should be created for each request. Otherwise, we immediately will have serious problems that are hard to debug

src/main/java/io/tarantool/driver/api/TarantoolCallOperations.java
akudiyar: Looking at this again, I strongly believe that we need to have this factory method in the ResultMapperFactoryFactory. Otherwise it looks like an orphan

ArtDu: > Looking at this again, I strongly believe that we need to have this factory method in the ResultMapperFactoryFactory. Otherwise it looks like an orphan

I don't want to see this, it's awful 
```java
client.getResultMapperFactoryFactory().getTarantoolTupleResultMapperFactory()
            .withSingleValueRowsMetadataToTarantoolTupleResultMapper
```

akudiyar: C'est la vie \_o_/

It's not used everyday, so an orphan method in API will confuse the users more. It's a classical leaking abstraction.

ArtDu: > It's not used everyday, so an orphan method in API will confuse the users more

It's just a shortcut for main mappers. But I'll move if you don't change your mind about using it here where is it now. I need it to futher autoTupleConverter

In my opinion, it's hard to use from `client.getResultMapperFactoryFactory().*` cause there are so many methods. It's so difficult for people that I suggeseted to use these methods. Every time I started to think how does these ugly structures. And the sum thing that these structure don't give a perfomance in comparing to http because we have a lot of problems with those.

@Elishtar What do you think about it? I think we need third opinion

akudiyar: You are leaking a concrete `TarantoolTuple` implementation to the `TarantoolClient` interface which has this type as a generic type. We should not do this by any chance. It harms more than the "economy" of symbols in one place or the "ugliness" of this factory name, That's the complexity of the source object that we are modeling and we need to get it away from the users by providing non-confusing and clear structures to them.

akudiyar: It's better to remove the method `getTarantoolTupleResultMapperFactory()` itself and use this factory as a singleton than to add a new method that is necessary only for internal purposes. So I'd vote for removing `getTarantoolTupleResultMapperFactory()` from the `TarantoolClient` interface instead

Elishtar: @akudiyar for me `ResultMapperFactoryFactory` looks very strange and over complexity, maybe better use few different factories?

akudiyar: @Elishtar the initial goal of this factory was to provide helper factory methods for the common cases like this one (a concrete tuple implementation + a concrete result type). Eventually these factory methods were intended to return singleton references. Changing this to individual factory/singleton variable instantiations may work, but to me looks like less readable and harder to track and alter the places of factories usages.

src/main/java/io/tarantool/driver/core/AbstractTarantoolClient.java
ArtDu: Moved factory to ResultMapperFactoryFactory

src/main/java/io/tarantool/driver/api/TarantoolCallOperations.java
ArtDu: We discussed that we move this factory to ResultMapperFactoryFactory now and that we should discuss API in more detailed later

akudiyar (approved):
LGTM

PR #324, opened 2022-12-13
Add separate metadata converters

### LINKED ISSUE #325: Use metadata parsers for TarantoolSpaceMetadata instead of converters
We have duplication code in DDL, _VSpace, RowsMetadata converters. We can move metadata parsing to the different package

ArtDu:
Changes reduced code duplication and helped to use it easily 

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes #325


src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+
+/**
+ * Populates space metadata. Input is expected to have the format which is returned by DDL module in spaces section.
+ * See
+ * <a href="https://github.com/tarantool/ddl#input-data-format">https://github.com/tarantool/ddl#input-data-format</a>
+ *
+ * @author Artyom Dubinin
+ */
+public final class DDLSpaceMetadataParser extends BaseMetadataParser {
```
akudiyar, line 18: I don't get the meaning of having an inheritance between classes only with static methods. I believe it will be better to have a normal class taking an `ArrayValue` in the constructor and with different `parseSpaceFormat` methods

src/main/java/io/tarantool/driver/core/metadata/parsers/VSpaceMetadataParser.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Populates space metadata. Input is expected to have the _vspace format.
+ * See
+ * <a href="https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_vspace/">
+ * https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/_vspace/
+ * </a>
+ *
+ * @author Artyom Dubinin
+ */
+public final class VSpaceMetadataParser extends BaseMetadataParser {
```
akudiyar, line 21: The same comment applies here as for DDLSpaceMetadataParser

akudiyar (changes requested):
We need unit tests for all three `parsingSpaceFormat` methods. Relying only on the integration tests is not productive since we cannot and shouldn't cover all the cases with them.

src/main/java/io/tarantool/driver/api/metadata/TarantoolSpaceMetadata.java
```diff
@@ -64,4 +70,16 @@ public interface TarantoolSpaceMetadata extends Serializable {
      * @return field name or null if this field not found in format metadata
      */
     Optional<String> getFieldNameByPosition(int fieldPosition);
+
+    static TarantoolSpaceMetadata parseDDLSpace(Map.Entry<Value, Value> space) {
```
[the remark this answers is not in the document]
ArtDu: Why do we need a factory? It's more difficult then we have static factory methods here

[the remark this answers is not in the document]
ArtDu: > I don't get why should it be in the TarantoolSpaceMetadata interface as a static method? It's not a part of API, it's something internal.

We can move it to TarantoolSpaceMetadataImpl

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
ArtDu: These classes don't store anything. They're only needed for creating a different classes like factories. I don't mind creating reusable *Parser like a normal class. We can store them in TarantoolSpaceMetadataImpl. The one problem we have with this approach that we have 
unnecessary new call(creation) every time when we create TarantoolSpaceMetadata.

src/main/java/io/tarantool/driver/core/metadata/TarantoolFieldMetadataImpl.java
```diff
@@ -39,7 +39,7 @@ public TarantoolFieldMetadataImpl(String fieldName, String fieldType, int fieldP
      * @param fieldPosition field position in tuple starting from 0
      * @param isNullable    is field nullable
      */
-    TarantoolFieldMetadataImpl(String fieldName, String fieldType, int fieldPosition, boolean isNullable) {
+    public TarantoolFieldMetadataImpl(String fieldName, String fieldType, int fieldPosition, boolean isNullable) {
```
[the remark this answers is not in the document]
ArtDu: Fixed

ArtDu:
@akudiyar I've added unit tests

akudiyar (changes requested):
The tests look OK (although a bit minimalistic), but I still don't see much sense in having separate stateless parser classes with both inheritance and static methods instead of encapsulating the parsing logic in the inherited special converters.  

Let's discuss this in PM

akudiyar (approved):
LGTM

src/main/java/io/tarantool/driver/core/metadata/parsers/DDLSpaceMetadataParser.java
akudiyar: We discussed this and decided to use our converters concept for the metadata parsing.

PR #330, opened 2022-12-19
Use options like mixins in proxy operations

### LINKED ISSUE #331: Problem with code duplication in ProxyClient options
We need it to avoid duplication of code when we need have something like below:
image
In this picture we need to use multiply inheritance, but it doesn't work with java

ArtDu:
We need it to avoid duplication of code when we need have something like below:
image
In this picture we need to use multiply inheritance, but it doesn't work with java

before:
image

after:
image

I haven't forgotten about:
- [ ] Tests
- [ ] Changelog (It doesn't change API)
- [ ] Documentation

Related issues:
Closes #331


ArtDu:
@akudiyar @Elishtar Also you can see how it looks with another option. Spoiler - it looks awful
https://github.com/tarantool/cartridge-java/pull/263

akudiyar (changes requested):
I remember why we went with the code duplication option instead of this way, although it seems appealing. The problem is that this implementation is OK for the proxy (CRUD) operations, and is not correct for the box (single instance) operations, because they support neither bucket ID nor timeout. In this case the user will see "withBucketId" option in the options builder, for example, and will be confused because it doesn't have any sense for the box operation. In that case, we need to fire an exception when the user calls an unsupported "add" method for making the user aware that that particular client type does not support it.

The best place for such an exception is when a call request is being built and we are passing the options to it. We need to have the corresponding unit tests as well.

akudiyar (approved):

ArtDu:
> UPD: nvm. THe old approach had the same problem with bucket ID available for the single instance client.

Yeah, we can fix it by splitting `SelectOption` to `SelectOption` and `ProxySelectOption` interfaces

PR #335, opened 2022-12-28
Batch operation errors parsing

### LINKED ISSUE #334: Throw an exception if insertMany or ReplaceMany batch operations call returned any errors
These batch operations do not return exceptions in a normal "nil, error" format, but rather return "result, errors" that contains both results of the successful tuple inserts/replaces and errors for the unsuccessful ones. Although this approach is questionable, we have no other options now than to support this exotic return format.

I suggest throwing the first returned error because we can not do anything with the other errors rather than retrying the operation completely. Otherwise, the cluster will still be in an unknown inconsistent state.

akudiyar:
I decided to not make a separate `SingleValueCallResult` parser for batch operations specifically because I believe that the return format must be fixed in tarantool/crud and it will change in the future. So I added a new case to the common parser and refactored it a bit.

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation
- [ ] Commit messages comply with the [guideline](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message)

Related issues:
Fixes #334

### DIFF
--- a/src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
+++ b/src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
@@ -8,6 +8,8 @@ import io.tarantool.driver.mappers.converters.ValueConverter;
 import org.msgpack.value.ArrayValue;
 import org.msgpack.value.Value;
 
+import java.util.function.Function;
+
 /**
  * Basic {@link SingleValueCallResult} implementation. If the result array contains two values where the first is
  * {@code null}, the second is treated as a formatted error or an error message.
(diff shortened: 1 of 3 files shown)


src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
```diff
@@ -20,34 +23,41 @@
...
+        if (callResultSize == 0 || result.size() == 1 && resultValue.isNilValue()) {
             // [nil] or []
-            value = null;
-        } else if (result.size() == 2 && (result.get(0).isNilValue() && !result.get(1).isNilValue())) {
+            return null;
+        } else if (callResultSize == 2 && (resultValue.isNilValue() && !errorsValue.isNilValue())) {
             // [nil, "Error msg..."] or [nil, {str="Error msg...", stack="..."}]
-            throw TarantoolErrorsParser.parse(result.get(1));
+            throw TarantoolErrorsParser.parse(errorsValue);
+        } else if (callResultSize == 2 && errorsValue.isArrayValue()) {
```
ArtDu, line 46: nit: You don't need else for every conditional branch here because you return or throw in them

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
```diff
@@ -20,34 +23,41 @@
...
+    }
+
+    private T parseResult(ArrayValue result, Function<Value, T> valueGetter) {
         if (result == null) {
             throw new TarantoolFunctionCallException("Function call result is null");
         }
-        if (result.size() == 0 || result.size() == 1 && result.get(0).isNilValue()) {
+        int callResultSize = result.size();
+        Value resultValue = result.getOrNilValue(0);
+        Value errorsValue = result.getOrNilValue(1);
```
ArtDu, line 39: Does it change anything? Below you use the same `isNilValue()`

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
```diff
@@ -20,34 +23,41 @@
...
             // [nil] or []
-            value = null;
-        } else if (result.size() == 2 && (result.get(0).isNilValue() && !result.get(1).isNilValue())) {
+            return null;
+        } else if (callResultSize == 2 && (resultValue.isNilValue() && !errorsValue.isNilValue())) {
             // [nil, "Error msg..."] or [nil, {str="Error msg...", stack="..."}]
-            throw TarantoolErrorsParser.parse(result.get(1));
+            throw TarantoolErrorsParser.parse(errorsValue);
+        } else if (callResultSize == 2 && errorsValue.isArrayValue()) {
+            // [result, nil] or [result, errors]
```
ArtDu, line 47: You checked that second argument is not nil
```suggestion
            // [result, []] or [result, errors]
```

ArtDu (changes requested):

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
akudiyar: Fixed the comment, thanks

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
akudiyar: It doesn't change the logic, although saves a few CPU cycles

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
akudiyar: Let's leave it as is for verbosity and logic visibility

src/main/java/io/tarantool/driver/core/SingleValueCallResultImpl.java
akudiyar: getOrNilValue() returns NilValue in case if the index is out of range.

ArtDu (approved):

Issue #117, opened 2021-08-20
Implement mapper caching to reduce GC utilization.


vrogach2020:
subj

vrogach2020:
Also we must consider serialization code in springdata and provide perfomance

ArtDu:
@akudiyar thoughts

>Raw idea: we should define a "tagging" API for mappers (e.g. a hashcode for the whole stack) that we can use for caching >the other mappers. Using such tag (hashcode) can allow caching mappers without space schema involved, but it will >require each converter and mapper to provide a unique tag (hashcode).

PR #360, opened 2023-02-17
Add deep copy instead of shallow copy in mappers

### LINKED ISSUE #166: No way to create a modifiable copy of DefaultMessagePackMapper. Now the copy affects original object
The case is: we want to create a copy of the default mapper and add some custom converters there. But we don't want to affect all existing converter stack. For example - I want to create a method that converts all Integers to Strings for some specific space..

We can't reuse a custom `MessagePackMapper` right now.

https://github.com/tarantool/cartridge-java/blob/5ed7e3e7f717d0c0c1347a58a0fdd9ac3996371b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java#L293-L296
https://github.com/tarantool/cartridge-java/blob/5ed7e3e7f717d0c0c1347a58a0fdd9ac3996371b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java#L43-L53
https://github.com/tarantool/cartridge-java/blob/5ed7e3e7f717d0c0c1347a58a0fdd9ac3996371b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java#L27-L31
 
`tarantoolClient.getConfig().getMessagePackMapper().copy();`
Deepcopy is not used here, so copying does not make sense. It only copies the surface layer:
if there was an entry in `tarantoolClient.getConfig().getMessagePackMapper()` in `Map<String, List<ValueConverter<? extends Value, ?>>> valueConverters; ` with the key `org.msgpack.value.ArrayValue`, then the copied list `List<ValueConverter<? extends Value, ?>` will be the same for all copied configs, and when registering a new converter for the copied config, it will appear for the original one as well.

For details see these comments' threads:
https://github.com/tarantool/cartridge-springdata/pull/80#discussion_r794516756
https://github.com/tarantool/cartridge-springdata/pull/80#discussion_r802047460
https://github.com/tarantool/cartridge-springdata/pull/80#discussion_r802056703

ArtDu:
I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes #166

### DIFF
--- a/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
+++ b/src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static io.tarantool.driver.mappers.MapperReflectionUtils.getInterfaceParameterClass;
 
@@ -54,9 +55,10 @@ public class DefaultMessagePackMapper implements MessagePackMapper {
      * @param mapper another mapper instance
      */
     public DefaultMessagePackMapper(DefaultMessagePackMapper mapper) {
-        this();
-        this.valueConverters.putAll(mapper.valueConverters);
-        this.objectConverters.putAll(mapper.objectConverters);
+        this.valueConverters = mapper.valueConverters.entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, e -> new ArrayList<>(e.getValue())));
+        this.objectConverters = mapper.objectConverters.entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getKey, e -> new ArrayList<>(e.getValue())));
     }
 
     @SuppressWarnings("unchecked")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 - Close public access to TarantoolResult*Impl ([#326](https://github.com/tarantool/cartridge-java/issues/326))
+- Add deep copy instead of shallow copy in default message pack mapper
+
 ## [0.10.1] - 2023-01-13
 
 ### Internal and API changes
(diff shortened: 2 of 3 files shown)


src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
```diff
@@ -54,9 +55,10 @@ public DefaultMessagePackMapper() {
      * @param mapper another mapper instance
      */
     public DefaultMessagePackMapper(DefaultMessagePackMapper mapper) {
-        this();
-        this.valueConverters.putAll(mapper.valueConverters);
-        this.objectConverters.putAll(mapper.objectConverters);
+        this.valueConverters = mapper.valueConverters.entrySet().stream()
```
akudiyar, line 58: Although it's a good idea on its own, it will lead to large memory consumption growth. Remember that we are currently creating a mapper stack on each request.

Did this change come out of a recently encountered bug? I was worried about that after we made the last mappers refactoring, see my comment https://github.com/tarantool/cartridge-java/pull/300#pullrequestreview-1203576211. Maybe let's think about proper mapper caching this time?

Raw idea:  we should define a "tagging" API for mappers (e.g. a hashcode for the whole stack) that we can use for caching the other mappers. Using such tag (hashcode) can allow caching mappers without space schema involved, but it will require each converter and mapper to provide a unique tag (hashcode).

akudiyar (changes requested):

src/main/java/io/tarantool/driver/mappers/DefaultMessagePackMapper.java
ArtDu: > Did this change come out of a recently encountered bug?

It's problem in springdata, because there we use registerValueConverter per request there. We should also think about how to separate custom mappers from default ones besides caching

ArtDu: I added your thoughts in https://github.com/tarantool/cartridge-java/issues/117

And also created a ticket about my separation idea https://github.com/tarantool/cartridge-java/issues/361

Elishtar (approved):

PR #362, opened 2023-02-27
Add default client request timeout

### LINKED ISSUE #87: If an address provider returns null, it produces NPE when establishing connections
Although this exception is then wrapped in the TarantoolConnectionException, there should be more clear exception anyway rather then a rough NPE:

```
Caused by: io.tarantool.driver.exceptions.TarantoolConnectionException: The client is not connected to Tarantool server
	at io.tarantool.driver.core.AbstractTarantoolConnectionManager.lambda$getConnection$0(AbstractTarantoolConnectionManager.java:96) ~[cartridge-driver-0.4.1.jar!/:na]
	at java.base/java.util.concurrent.CompletableFuture.uniHandle(Unknown Source) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture.uniHandleStage(Unknown Source) ~[na:na]
	at java.base/java.util.concurrent.CompletableFuture.handle(Unknown Source) ~[na:na]
	at io.tarantool.driver.core.AbstractTarantoolConnectionManager.getConnection(AbstractTarantoolConnectionManager.java:88) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.AbstractTarantoolClient.makeRequest(AbstractTarantoolClient.java:411) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.AbstractTarantoolClient.call(AbstractTarantoolClient.java:228) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.AbstractTarantoolClient.call(AbstractTarantoolClient.java:222) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.AbstractTarantoolClient.call(AbstractTarantoolClient.java:216) ~[cartridge-driver-0.4.1.jar!/:na]
	... 33 common frames omitted
Caused by: java.lang.NullPointerException: null
	at io.tarantool.driver.core.AbstractTarantoolConnectionManager.getConnections(AbstractTarantoolConnectionManager.java:175) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.core.AbstractTarantoolConnectionManager.establishConnections(AbstractTarantoolConnectionManager.java:160) ~[cartridge-driver-0.4.1.jar!/:na]
	at io.tarantool.driver.core.AbstractTarantoolConnectionManager.getConnectionInternal(AbstractTarantoolConnectionManager.java:112) ~[cartridge-driver-0.4.1.jar!/:na]
	... 38 common frames omitted
```

ArtDu:
@ akudiyar changed the behaviour https://github.com/tarantool/cartridge-java/commit/57af227042cdc6402749e39a70a49ae61b94eb9a#diff-b2ae385ada02a75ad946141c7af41540970760d41d02d54209c00eabdc9bc8e3L175

I suggest make "OR" condition. If we have options.timeout, we use it, otherwise we use client.getTimeout()

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:

### DIFF
--- a/src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceDeleteOptionsIT.java
+++ b/src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceDeleteOptionsIT.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 
+import static org.junit.Assert.assertNull;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
@@ -87,7 +88,11 @@ public class ProxySpaceDeleteOptionsIT extends SharedCartridgeContainer {
         Conditions conditions = Conditions.equals(PK_FIELD_NAME, 1);
         profileSpace.delete(conditions).get();
         List<?> crudDeleteOpts = client.eval("return crud_delete_opts").get();
-        assertEquals(requestConfigTimeout, ((HashMap) crudDeleteOpts.get(0)).get("timeout"));
+        assertNull(((HashMap) crudDeleteOpts.get(0)).get("timeout"));
+
+        profileSpace.delete(conditions, ProxyDeleteOptions.create()).get();
+        crudDeleteOpts = client.eval("return crud_delete_opts").get();
+        assertNull(((HashMap) crudDeleteOpts.get(0)).get("timeout"));
 
         // with option timeout
         profileSpace.delete(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- Remove config request timeout as default parameter in crud client
 - Close public access to TarantoolResult*Impl ([#326](https://github.com/tarantool/cartridge-java/issues/326))
 - Add deep copy instead of shallow copy in default message pack mapper
 
--- a/src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java
+++ b/src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java
@@ -78,9 +78,7 @@ public abstract class ProxyTarantoolSpace<T extends Packable, R extends Collecti
 
     @Override
     public CompletableFuture<R> delete(Conditions conditions) throws TarantoolClientException {
-        return delete(conditions, rowsMetadataTupleResultMapper(), ProxyDeleteOptions.create()
-            .withTimeout(config.getRequestTimeout())
-        );
+        return delete(conditions, rowsMetadataTupleResultMapper(), ProxyDeleteOptions.create());
     }
 
     @Override
@@ -113,9 +111,7 @@ public abstract class ProxyTarantoolSpace<T extends Packable, R extends Collecti
 
     @Override
     public CompletableFuture<R> insert(T tuple) throws TarantoolClientException {
-        return insert(tuple, rowsMetadataTupleResultMapper(), ProxyInsertOptions.create()
-            .withTimeout(config.getRequestTimeout())
-        );
+        return insert(tuple, rowsMetadataTupleResultMapper(), ProxyInsertOptions.create());
     }
 
     @Override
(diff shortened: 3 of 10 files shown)


src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceDeleteOptionsIT.java
```diff
@@ -89,6 +89,10 @@ public void withTimeout() throws ExecutionException, InterruptedException {
         List<?> crudDeleteOpts = client.eval("return crud_delete_opts").get();
         assertEquals(requestConfigTimeout, ((HashMap) crudDeleteOpts.get(0)).get("timeout"));
 
+        profileSpace.delete(conditions, ProxyDeleteOptions.create()).get();
+        crudDeleteOpts = client.eval("return crud_delete_opts").get();
+        assertEquals(requestConfigTimeout, ((HashMap) crudDeleteOpts.get(0)).get("timeout"));
```
akudiyar, line 94: Seems that setting the request timeout is missing here

akudiyar: That also applies to all the other tests

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
```diff
@@ -49,8 +50,9 @@ public Builder<T> withIndexQuery(TarantoolIndexQuery indexQuery) {
         }
 
         public DeleteProxyOperation<T> build() {
+            Optional optionsTimeout = options.getTimeout();
             CRUDBucketIdOptions requestOptions = new CRUDDeleteOptions.Builder()
-                .withTimeout(options.getTimeout())
+                .withTimeout(optionsTimeout.isPresent() ? optionsTimeout : requestTimeout)
```
akudiyar, line 55: I'm not sure that we need two ways of setting the timeout. For what the other timeout (in options) then stands for? In the previous variant, the request timeout was passed to the response handler. I believe that we need to make them distinct timeouts and pass both of them in the options.

akudiyar: Passing in the options is optional, but passing to the handler is mandatory I think

akudiyar (changes requested):

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
ArtDu: But the main think about the changes that we have passing config.requestTimeout https://github.com/tarantool/cartridge-java/blob/97f2a19a4e90b90a7ffa515d4175a5daec10fa2c/src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java#L251-L256

For example if we used fields options we wouldn't have this feature and it might frustrate. This is the thing that @vlalykin complained:

> добрый день.
по поводу возможности задания списка полей в options: https://github.com/tarantool/cartridge-java/issues/236

>если я задаю новый параметр для уточнения списка полей, то не передается дефолтный таймаут (который раньше задавался один раз на уровне клиента). теперь его надо каждый раз руками задавать, или я что то неправильно делаю ?

>client.getSpace(SPACE).select(condition, ProxySelectOptions.create().withFields(fields))

akudiyar: The timeout that is passed currently to the options is different. We need to not pass the `requestTimeout` instead of that timeout, but rather pass this `requestTimeout` to the request handler as it was done previously (that's what the users want).

akudiyar (changes requested):
We need to return back the previous behavior, it is currently broken. The actual changes are doing something other.

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
ArtDu: We always pass requestTimeout in crud option. See [0.8.0](https://github.com/tarantool/cartridge-java/blob/fd90d07117fadbdf591af65404d3c793e306dbfb/src/main/java/io/tarantool/driver/core/space/ProxyTarantoolSpace.java#L139) version(it's before my options changes). Do you want to change this behavior?

akudiyar: Ah, it was broken even then. Yes, I think that we need to pass the `requestTimeout` to Netty, and only the timeout in the options to tarantool/crud.

akudiyar (changes requested):
We need to pass `requestTimeout` to the handlers and `timeout` in options to crud

Elishtar (approved):

akudiyar (approved):
LGTM, as it at least removes the ambiguity.

I've created a new issue https://github.com/tarantool/cartridge-java/issues/367 for implementing the `requestTimeout` option correctly.

PR #372, opened 2023-03-22
Add metadataFormatIsEmpty in TarantoolTuple

ArtDu:
It's needed when we expected crud and box responses by one mapper(AUTO mapper in cartridge-springdata). In case when we obtain box response and we didn't have prepared metadata from ddl/_vspace, we can't map these values to Entity without metadata and we can't check whether we have it or not. And we can't raise an exception that we don't have metadata before because we can take it from crud response. Springdata doesn't see a difference between crud and box responses, it only got TarantoolTupleResult

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Needed for https://github.com/tarantool/cartridge-springdata/issues/123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 - Remove config request timeout as default parameter in crud client
 - Add details for the case of space metadata fetching failure ([#200](https://github.com/tarantool/cartridge-java/issues/200))
+- Add metadataFormatIsEmpty in TarantoolTuple
 - Close public access to TarantoolResult*Impl ([#326](https://github.com/tarantool/cartridge-java/issues/326))
 - Split retrying into more detailed modules ([#341](https://github.com/tarantool/cartridge-java/issues/341))
 - Add deep copy instead of shallow copy in default message pack mapper ([#166](https://github.com/tarantool/cartridge-java/issues/166))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 - Remove config request timeout as default parameter in crud client
 - Add details for the case of space metadata fetching failure ([#200](https://github.com/tarantool/cartridge-java/issues/200))
-- Add metadataFormatIsEmpty in TarantoolTuple
+- Add hasMetadata in TarantoolTuple
 - Close public access to TarantoolResult*Impl ([#326](https://github.com/tarantool/cartridge-java/issues/326))
 - Split retrying into more detailed modules ([#341](https://github.com/tarantool/cartridge-java/issues/341))
 - Add deep copy instead of shallow copy in default message pack mapper ([#166](https://github.com/tarantool/cartridge-java/issues/166))
--- a/src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
+++ b/src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
@@ -15,6 +15,13 @@ import java.util.UUID;
  */
 public interface TarantoolTuple extends Iterable<TarantoolField>, Packable {
     /**
+     * Check whether tuple have information about space format
+     *
+     * @return true if we obtain fields by names
+     */
+    boolean metadataFormatIsEmpty();
+
+    /**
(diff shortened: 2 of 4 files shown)


src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
```diff
@@ -14,6 +14,13 @@
  * @author Alexey Kuzin
  */
 public interface TarantoolTuple extends Iterable<TarantoolField>, Packable {
+    /**
+     * Check whether tuple have information about space format
```
akudiyar, line 18:
```suggestion
     * Check whether this tuple includes information about space format
```

src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
```diff
@@ -14,6 +14,13 @@
...
+     *
+     * @return true if we obtain fields by names
```
akudiyar, line 20:
```suggestion
     * @return true if we can obtain the tuple fields by names
```

src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
```diff
@@ -14,6 +14,13 @@
...
+     */
+    boolean metadataFormatIsEmpty();
```
akudiyar, line 22: I suggest using a simpler naming like `hasMetadata` or `isMetadataEmpty`

akudiyar (changes requested):
This change is OK by itself, just some wording and naming needs to be fixed.

But there is a greater question, do we need this change at all. It seems it is not comfortable to work with a tuple that represents in fact two different entities.

Can it be better if derive a `TarantoolTupleWithMetadata` interface from the original `TarantoolTuple` one, move to the second one all methods that access fields by names, add the corresponding `*Impl` class, and then use that interface distinction in SpringData via `instanceof`? That will allow us to finally avoid all the checks for metadata presence that we have to use in current implementation, and the user confusion when they try to get a field by name and see a strange "metadata does not exist" error.
With that implementation, we can change the `*Proxy*` API classes to return `TarantoolTupleWithMetadata` tuples instead of `TarantoolTuple` ones. That may also probably require changing the `TarantoolTuple` interface somewhere either to `Packable` or to `? extends TarantoolTuple`.

src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
ArtDu: Renamed to `hasMetadata `

src/main/java/io/tarantool/driver/api/tuple/TarantoolTuple.java
```diff
@@ -14,6 +14,13 @@
  * @author Alexey Kuzin
  */
 public interface TarantoolTuple extends Iterable<TarantoolField>, Packable {
+    /**
+     * Check whether this tuple includes information about space format
+     *
+     * @return true if we can obtain the tuple fields by names
+     */
+    boolean hasMetadata();
```
akudiyar, line 22: If we are going to use different types (`TarantoolTuple` and `TarantoolTupleWithMetadata`). do we really need this method then? It is a public API, we need to carefully think before changing it

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 - Remove config request timeout as default parameter in crud client
 - Add details for the case of space metadata fetching failure ([#200](https://github.com/tarantool/cartridge-java/issues/200))
+- Add metadataFormatIsEmpty in TarantoolTuple
```
akudiyar, line 6: We need to change this entry too

akudiyar (changes requested):
I still have the same question, see below

CHANGELOG.md
ArtDu, line 6:
```suggestion
- Add hasMetadata in TarantoolTuple
```

akudiyar:
OK. Let's amend this change in the nearest updates then.

akudiyar (approved):

PR #373, opened 2023-03-22
Add can convert value to TupleResultConverter

ArtDu:
It's needed when we want to take maps in the array. Springdata can map map structures. Using AUTO converter we can't map a simple array without tuples to the array and pass conversion control to the next layer.

I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Needed for https://github.com/tarantool/cartridge-springdata/issues/123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Split retrying into more detailed modules ([#341](https://github.com/tarantool/cartridge-java/issues/341))
 - Add deep copy instead of shallow copy in default message pack mapper
 - Add a factory builder for constructing mapper hierarchies
+- Add can convert value to TupleResultConverter
 
 ## [0.10.1] - 2023-01-13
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Split retrying into more detailed modules ([#341](https://github.com/tarantool/cartridge-java/issues/341))
 - Add deep copy instead of shallow copy in default message pack mapper
 - Add a factory builder for constructing mapper hierarchies
+- Add can convert value to TupleResultConverter
```
akudiyar, line 9:
```suggestion
- Add "can convert value" check to TupleResultConverter
```

akudiyar (changes requested):
`TarantoolResult` is only about tuples. It's a special container class and it is necessary to distinct an array of arrays as tuples from an array of arrays that are not tuples or an array of any other objects (e.g. maps).
Would you mind explaining a bit more how this change will solve the original problem?

PR #383, opened 2023-04-14
Get rid of double executor in retrying

### LINKED ISSUE #382: Remove double executor run from RetryingClient.space
Now we have a performance problem so we run retrying wrapper in the executor and the retryable task in the executor

https://github.com/tarantool/cartridge-java/blob/d9ca853b6008137c2ea6c68a0246a68d43771d31/src/main/java/io/tarantool/driver/core/RetryingTarantoolClient.java#L444

ArtDu:
Before changing:
```jmh
Benchmark                   Mode  Cnt      Score       Error  Units
BenchmarkRunner.spaceCall  thrpt    5  74844.508 ± 16460.492  ops/s
```

After changing:
```jmh
Benchmark                   Mode  Cnt       Score        Error  Units
BenchmarkRunner.spaceCall  thrpt    5  275476.264 ± 256775.771  ops/s
```

* Error means mean margin between iteration:

> This is the margin of error for the score. In most cases, that is a half of [confidence interval](http://en.wikipedia.org/wiki/Confidence_interval). Think about it as if there is a "±" sign between "Score" and "Score error". In fact, the human-readable log will show that:

```
Iteration   1: 334564.445 ops/s
Iteration   2: 226016.924 ops/s
Iteration   3: 343260.667 ops/s
Iteration   4: 190627.608 ops/s
Iteration   5: 282911.676 ops/s
```

I haven't forgotten about:
- [ ] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes #382

### DIFF
--- a/src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
+++ b/src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
@@ -35,6 +35,7 @@ public class TarantoolSetup {
         .withLogConsumer(new Slf4jLogConsumer(log));
 
     TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>> tarantoolClient;
+    TarantoolClient<TarantoolTuple, TarantoolResult<TarantoolTuple>> retryingTarantoolClient;
     MessagePackMapper defaultMapper;
     CallResultMapper<TarantoolResult<TarantoolTuple>, SingleValueCallResult<TarantoolResult<TarantoolTuple>>>
         resultMapper;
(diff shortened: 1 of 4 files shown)


src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
```diff
@@ -49,6 +50,9 @@ private void initClient() {
             .withCredentials(tarantoolContainer.getUsername(), tarantoolContainer.getPassword())
             .build();
 
+        retryingTarantoolClient
+            = TarantoolClientFactory.configureClient(tarantoolClient).withRetryingByNumberOfAttempts(3).build();
```
vasilevichra, line 54: Such a long name. Why not `withRetry(3)` or `withAttempt(42)`?

vasilevichra (approved):

src/test/java/io/tarantool/driver/benchmark/TarantoolSetup.java
ArtDu: You're right, maybe we should rename it. Could you create a ticket for it?

akudiyar: This is a builder. We can make it hierarchical, but with plain structure as it is now such a long name is necessary because it specifies two options in one time: that we have a retry mechanism enabled and that we specify number of attempts as its parameter.

akudiyar (approved):
LGTM.

PR #391, opened 2023-05-02
Make mapper builder more universal

ArtDu:
I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:


src/main/java/io/tarantool/driver/api/TarantoolCallOperations.java
```diff
@@ -123,10 +123,10 @@ <T> CompletableFuture<TarantoolResult<T>> call(
      * @return some result
      * @throws TarantoolClientException if the client is not connected or some other error occurred
      */
-    <T> CompletableFuture<TarantoolResult<T>> call(
+    <T> CompletableFuture<T> call(
```
akudiyar, line 126: We will have to change the other `call()` functions that return `TarantoolResult` too - see lines 79, 91, 109

src/main/java/io/tarantool/driver/mappers/factories/ResultMapperFactoryFactory.java
```diff
@@ -359,8 +359,11 @@ Builder withArrayValueToTarantoolTupleResultConverter(
         Builder withRowsMetadataToTarantoolTupleResultConverter(
             MessagePackMapper messagePackMapper);
 
-        CallResultMapper buildCallResultMapper();
+        MessagePackValueMapper buildCallResultMapper();
 
-        CallResultMapper buildCallResultMapper(MessagePackMapper valueMapper);
+        MessagePackValueMapper buildCallResultMapper(MessagePackMapper valueMapper);
+
+        CallResultMapper<Object, SingleValueCallResult<Object>>
```
akudiyar, line 366: Here we should have generics (`T` and a parameter of type `Class<T>`) and not `Object`, that will be a very bad API otherwise

akudiyar, line 366: Here we should have generics (`T` and a parameter of type `Class<T>`) and not `Object`, that will be a very bad API otherwise

akudiyar, line 366: Here we should have generics (`T` and a parameter of type `Class<T>`) and not `Object`, that will be a very bad API otherwise

src/main/java/io/tarantool/driver/mappers/factories/ResultMapperFactoryFactory.java
```diff
@@ -359,8 +359,11 @@ Builder withArrayValueToTarantoolTupleResultConverter(
         Builder withRowsMetadataToTarantoolTupleResultConverter(
             MessagePackMapper messagePackMapper);
 
-        CallResultMapper buildCallResultMapper();
+        MessagePackValueMapper buildCallResultMapper();
```
akudiyar, line 362: Why do we need to change this?

akudiyar (changes requested):
Changing the `call` method signature seems reasonable, but that's a breaking change and we must notify the customers appropriately. Changing `MessagePackMapper` parameter types to `MessagePackValueMapper` also seems reasonable.

But we shouldn't have any raw  `Object`s in the API (`List<?>` in `call` and `eval` is the only exception here because of the compatibility with `tarantool-java` and also it's a better variant than just `Object`). ALso I don't get why we change `CallResultMapper` result types to something else.

src/main/java/io/tarantool/driver/mappers/factories/ResultMapperFactoryFactory.java
ArtDu: T doesn't fit for us. Because generic require us to detail type and it would be only one type. But I want to have mapper that can be fit for all possible responses. I had to use wildcard(?) for it like in generic call method, but there were problems as I remember. I trying to get off easy

src/main/java/io/tarantool/driver/api/TarantoolCallOperations.java
ArtDu: Changed

ArtDu:
> But we shouldn't have any raw `Object`s in the API (`List<?>` in `call` and `eval` is the only exception here because of the compatibility with `tarantool-java` and also it's a better variant than just `Object`). ALso I don't get why we change `CallResultMapper` result types to something else.

I added two ways of using with generics and without. We need object for universal mapping when return type is uncertain

src/main/java/io/tarantool/driver/mappers/factories/ResultMapperFactoryFactory.java
akudiyar: That contradicts with the method name then, and that means you need either a new method or you want some other change in behavior.

ArtDu: Fixed

ArtDu:
@akudiyar I probably did all what we discussed. And also I updated PR in springdata(https://github.com/tarantool/cartridge-springdata/pull/124). Please check both PR and give your opinion

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
```diff
@@ -442,33 +442,30 @@ public void test_universalConverter_shouldWorkWithBoxAndCrudCorrectly() {
...
-            .withSingleValueConverter(
-                factory.createMapper(valueMapper)
-                    .withArrayValueToTarantoolTupleResultConverter()
-                    .withRowsMetadataToTarantoolTupleResultConverter()
-                    .buildCallResultMapper()
-            )
-            .buildCallResultMapper();
-
-        Object crudResult =
+        CallResultMapper<TarantoolTupleResultImpl, SingleValueCallResult<TarantoolTupleResultImpl>> callReturnMapper =
```
akudiyar, line 445: We shouldn't have `Impl` classes normally in the generic type spec. Is it intentional or a mistype?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Internal and API changes
+- **[breaking change]** Split call API for calls with tarantool result as return and call with generic return ([#123](https://github.com/tarantool/cartridge-springdata/issues/123))
```
akudiyar, line 6:
```suggestion
- **[breaking change]** Split call API into calls with tuple result as return type and calls with generic Object return type ([#123](https://github.com/tarantool/cartridge-springdata/issues/123))
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+- Add buildSingleValueResultMapper in mapper builder
```
akudiyar, line 7:
```suggestion
- Add buildSingleValueResultMapper() in mapper builder for generic call result mappers. That will simplify the support for auto object converter in SpringData.
```

akudiyar (approved):
That's in the much better shape now. I have just three small nitpicks

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
ArtDu: We have to pass class as argument of method. And I didn't want to use ugly casts (class<TarantoolResult<TarantoolTuple>>) (class<?>) anyClass

ArtDu: Oh, I got what you meant. I'll fix it now

akudiyar (approved):
LGTM

Issue #404, opened 2023-07-14
Refresh connections isn't working if restart is faster than discovery task 


ArtDu:
It's happened because if the refresh callback isn't called if the discovery task returns the same address list

https://github.com/tarantool/cartridge-java/blob/67c9f67a34872ed0de70044d85fa2476ef78a3ec/src/main/java/io/tarantool/driver/cluster/AbstractDiscoveryClusterAddressProvider.java#L36-L53

PR #405, opened 2023-07-25
Fix losing connection if restart happened faster than discovery task

ArtDu:
I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes https://github.com/tarantool/cartridge-java/issues/404

### DIFF
--- a/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
+++ b/src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
@@ -17,6 +17,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -93,6 +94,25 @@ public abstract class AbstractTarantoolConnectionManager implements TarantoolCon
     @Override
     public boolean refresh() {
         return connectionMode.compareAndSet(ConnectionMode.OFF, ConnectionMode.PARTIAL);
+    }
+
+    protected boolean areAddressesChanged() {
+        Collection<TarantoolServerAddress> addresses = getAddresses();
+        if (addresses == null) {
+            logger.debug("The list of server addresses is not defined");
+            return true;
+        }
+        return !connectionRegistry.keySet().equals(new HashSet<>(addresses));
+    }
+
+    protected boolean areConnectionsAlive() {
+        for (List<TarantoolConnection> connections : connectionRegistry.values()) {
+            int isAliveConnections = (int) connections.stream().filter(TarantoolConnection::isConnected).count();
+            if (isAliveConnections != config.getConnections()) {
+                return false;
+            }
+        }
+        return true;
     }
 
     private CompletableFuture<TarantoolConnection> getConnectionInternal() {
--- a/src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
+++ b/src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
@@ -1,5 +1,30 @@
 package io.tarantool.driver.integration;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.org.apache.commons.lang3.reflect.FieldUtils;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import io.tarantool.driver.TarantoolUtils;
 import io.tarantool.driver.api.SingleValueCallResult;
 import io.tarantool.driver.api.TarantoolClient;
 import io.tarantool.driver.api.TarantoolClientConfig;
@@ -27,7 +52,7 @@ import io.tarantool.driver.cluster.TestWrappedClusterAddressProvider;
 import io.tarantool.driver.core.ClusterTarantoolTupleClient;
 import io.tarantool.driver.core.ProxyTarantoolTupleClient;
 import io.tarantool.driver.core.RetryingTarantoolTupleClient;
-import io.tarantool.driver.core.TarantoolTupleResultImpl;
+import io.tarantool.driver.core.connection.TarantoolClusterConnectionManager;
 import io.tarantool.driver.core.tuple.TarantoolTupleImpl;
 import io.tarantool.driver.exceptions.TarantoolInternalException;
 import io.tarantool.driver.exceptions.TarantoolNoSuchProcedureException;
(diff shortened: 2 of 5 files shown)


src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -92,9 +94,29 @@ public CompletableFuture<TarantoolConnection> getConnection() {
 
     @Override
     public boolean refresh() {
+        if (areConnectionsAlive()) {
```
akudiyar, line 97: I think that is not useful - we may want to trigger restart even if (all or some of) the connections are dead (like, the Tarantool server is restarting while the application with driver is still online).

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -92,9 +94,29 @@ public CompletableFuture<TarantoolConnection> getConnection() {
...
+            return false;
+        }
+
+        for (List<TarantoolConnection> connections : connectionRegistry.values()) {
+            int isAliveConnections = (int) connections.stream().filter(TarantoolConnection::isConnected).count();
+            if (isAliveConnections != config.getConnections()) {
+                return false;
+            }
+        }
+        return connectionRegistry.keySet().equals(new HashSet<>(addresses));
```
akudiyar, line 117: What is the meaning of this check?

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
```diff
@@ -1,5 +1,30 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.org.apache.commons.lang3.reflect.FieldUtils;
```
akudiyar, line 20: That's a bad import, do we really need it?

akudiyar (changes requested):

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
ArtDu: I think it's OK for testing purposes

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: You comment is reasonable. I'll think about it. I guess we have to check connections outside refresh. Without it discovery is broken. 

Or we can always run refresh in every discovery task. I haven't followed this approach because in each refresh we use mutex(phaser). And it probably follows problem with less perfomance

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: If addreses have changed we need to drop the old ones. I moved this logic from discovery tssk code

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
akudiyar: I'd rather avoid such import even in the tests, that pollutes the library classpath. Import the commons library (or even a better one), but don't use an internal dependency of a dependency

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
akudiyar: The refresh trigger was designed to be used in the discovery task explicitly. So if it is called independently somewhere before the discovery returns any addresses, that is incorrect.

akudiyar: Looks like we introduced a `refresh` call somewhere and forgot about its purpose

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
```diff
@@ -672,4 +680,63 @@ public void multiResultWithErrorTest() throws ExecutionException, InterruptedExc
...
+            } catch (IllegalAccessException | NoSuchMethodException e) {
+                throw new RuntimeException(e);
+            }
+        });
+    }
+
+    private static int getAliveConnections() throws IllegalAccessException, NoSuchMethodException {
+        if (addressProvider == null) {
+            ClusterTarantoolTupleClient
+                innerClient = (ClusterTarantoolTupleClient) FieldUtils.readField(
```
akudiyar, line 712: If we need to test this field (do we need really?), let's provide an (internal) interface to retrieve it. Using reflection is unnecessary.

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
ArtDu: Aa, you are about import string. I'll change it

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: > The refresh trigger was designed to be used in the discovery task explicitly. So if it is called independently somewhere before the discovery returns any addresses, that is incorrect.

Yeap, I changed it. Now refresh is like it was before. Discovery calls different function that checks conditions and calls refresh

src/test/java/io/tarantool/driver/integration/ProxyTarantoolClientIT.java
ArtDu: In this commit, I've shown what we have to add to check real connections via the client 5534b16a8aee3a129bae8c0c6d44a76ae3dda730

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -90,11 +91,33 @@ public CompletableFuture<TarantoolConnection> getConnection() {
...
+    public Map<TarantoolServerAddress, List<TarantoolConnection>> getConnectionRegistry() {
+        return connectionRegistry;
+    }
+
     @Override
     public boolean refresh() {
         return connectionMode.compareAndSet(ConnectionMode.OFF, ConnectionMode.PARTIAL);
     }
 
+    protected boolean areConditionsChanged() {
```
iDneprov, line 103: Update method logics

ArtDu: Fixed

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
```diff
@@ -90,11 +91,33 @@ public CompletableFuture<TarantoolConnection> getConnection() {
         });
     }
 
+    public Map<TarantoolServerAddress, List<TarantoolConnection>> getConnectionRegistry() {
```
akudiyar, line 94: Should it be public really? Or package-private is enough?

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
akudiyar, line 103: Let's rename that method to something more specific. Like `haveAddressesChanged`

src/main/java/io/tarantool/driver/core/RetryingTarantoolClient.java
```diff
@@ -93,6 +94,11 @@ public TarantoolVersion getVersion() throws TarantoolClientException {
         return client.getVersion();
     }
 
+    @Override
+    public TarantoolConnectionManager getConnectionManager() {
```
akudiyar, line 98: Should it be public? Make it package-private if possible

akudiyar (changes requested):
I have a few questions, otherwise looks good

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: Public because we call this method in test

src/main/java/io/tarantool/driver/core/RetryingTarantoolClient.java
ArtDu: It should be because we hold TarantoolClient interface in Proxy/Retrying clients as inner client

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: I split the method on two

```
areAddressesChanged
```

```
areConnectionsAlive
```

src/main/java/io/tarantool/driver/core/connection/AbstractTarantoolConnectionManager.java
ArtDu: I realized that we have connection listeners and we don't need to expose connections itself

akudiyar (approved):

PR #416, opened 2023-09-05
Added "mode" option for crud select operation

### LINKED ISSUE #107: Implement API to pass 'mode=write' parameter to CRUD library
Sometimes it's necessary to read from 'master' node to ensure that the fetched data is up-to date.
See [select opts: mode=write](https://github.com/tarantool/crud).

Also it would be nice to have some generic way to pass parameters to CRUD or via springdata methods.

nickkkccc:
I haven't forgotten about:
- [x] Tests
- [x] Changelog
- [ ] Documentation

Related issues:
Closes #107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### API changes
+- Add "mode" option for crud select operation ([#107](https://github.com/tarantool/cartridge-java/issues/107))
+
 ### Bugfixes
 - Fix Instant converter to parse 8 bytes datetime ([#408](https://github.com/tarantool/cartridge-java/issues/408))
 
--- a/src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
+++ b/src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
@@ -1,0 +1,38 @@
+package io.tarantool.driver.api.space.options;
+
+import java.util.Optional;
+
+/**
+ * Base interface for all operation options that may have a configurable mode.
+ *
+ * @author Belonogov Nikolay
+ */
+public interface OperationWithModeOptions<T extends OperationWithModeOptions<T>> extends Options, Self<T> {
+    String MODE = "mode";
+
+    /**
+     * Specifies the mode for operations (select, count, get) on a specific node type (mode == "write" - master, mode
+     * == "read" - replica). By default, mode is "read".
+     *
+     * @param mode mode for operations (select, get, count).
+     * @return this options instance.
+     */
+    default T withMode(String mode) {
+        if (!mode.equals("read") && !mode.equals("write")) {
+            throw new IllegalArgumentException("Mode should be \"read\" or \"write\"");
+        }
+
+        addOption(MODE, mode);
+        return self();
+    }
+
+    /**
+     * Return operation mode.
+     *
+     * @return mode.
+     */
+    default Optional<String> getMode() {
+        return getOption(MODE, String.class);
+    }
+
+}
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### API changes
```
ArtDu, line 5: It's better to squash all these commits to one

src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Specifies the mode for operations (select, count, get) on a specific node type (mode == "write" - master, mode
+     * == "read" - replica). By default, mode is "read".
+     *
+     * @param mode mode for operations (select, get, count).
+     * @return this options instance.
+     */
+    default T withMode(String mode) {
+        if (!mode.equals("read") && !mode.equals("write")) {
```
ArtDu, line 21: nit: You can use enum because modes are determined values. It's up to you

ArtDu (commented):
Thanks for the PR. Great job, I left only two small comments

src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
akudiyar: I suggest using enums. This check will look better then. Also, the users will benefit from using the enum values that are better supported by autocompletion in editors.

src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceSelectOptionsIT.java
```diff
@@ -167,4 +152,23 @@ public void withFieldsTest() throws ExecutionException, InterruptedException {
         assertEquals(1, tuple.getInteger("profile_id"));
         assertEquals(50, tuple.getInteger("age"));
     }
+
+    @Test
+    public void withMode() throws ExecutionException, InterruptedException {
```
akudiyar, line 157:
```suggestion
    public void withModeTest() throws ExecutionException, InterruptedException {
```

src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
akudiyar: Also I kinda not like the initial naming, it's rather strange. We may think about using something more widely used in the industry, like `leader`/`replica`, otherwise the meaning is very obscure (my first thought was "is `select` writing anything?!")

akudiyar (changes requested):
Looks good, I have two suggestions BTW

src/main/java/io/tarantool/driver/api/space/options/OperationWithModeOptions.java
ArtDu: > Also I do not like the initial naming, it's rather strange. We may think about using something more widely used in the industry, like `leader`/`replica`, otherwise the meaning is very obscure (my first thought was "is `select` writing anything?!")

It should have been renamed in crud/vshard. If we changed it in connector it would be confusuing thing for users who wants to understand how tarantool and ecosystem works. Let it be unified

nickkkccc: Thanks. Done.

src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceSelectOptionsIT.java
nickkkccc: Fix in commit.

ArtDu (approved):

PR #418, opened 2023-09-10
Make CRUDOptions unified

nickkkccc:
I haven't forgotten about:
- [ ] Tests
- [ ] Changelog
- [ ] Documentation
- [ ] Commit messages comply with the [guideline](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message)

Related issues:
Needed for #417


ArtDu (commented):
Speaking about enum and wrappers. 
First of all, it is a breaking change. 
The second one is why we use wrapper before enum. It makes code more complicated.
```diff
-.withMode(new ModeValue(ModeValue.WRITE))
+.withMode(ModeValue.WRITE)
```

I'm asking you to separate changes on two pull request:
1. "Make CRUDOptions use interface to reduce code duplications"
2. "Use enum for parameters"



.withMode(new ModeValue(ModeValue.WRITE))

src/main/java/io/tarantool/driver/api/space/options/crud/enums/CRUDOption.java
```diff
@@ -0,0 +1,32 @@
+package io.tarantool.driver.api.space.options.crud.enums;
+
+/**
+ * Base enum for all CRUD operation options.
+ *
+ * @author <a href="https://github.com/nickkkccc">Belonogov Nikolay</a>
+ */
+public enum CRUDOption {
```
ArtDu, line 8: We have the structure now where the CRUD prefix is used for inner interfaces, but the proxy prefix is used for user API

src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceUpdateOptionsIT.java
```diff
@@ -114,17 +116,17 @@ public void withBucketIdTest() throws ExecutionException, InterruptedException {
...
-        assertNull(((HashMap) crudUpdateOpts.get(0)).get("bucket_id"));
+        assertNull(((HashMap<?, ?>) crudUpdateOpts.get(0)).get(CRUDOption.BUCKET_ID.getName()));
 
         // with bucket id
         Integer bucketId = 1;
         profileSpace.update(
             conditions,
             tarantoolTuple,
-            ProxyUpdateOptions.create().withBucketId(bucketId)
+            ProxyUpdateOptions.create().withBucketId(new BucketIdValue(bucketId))
```
ArtDu, line 126: The problem with the wrapper is that we still pass just a value and validate it inside the call. The wrapper don't change pipeline, only add more words to pass the value

akudiyar: I agree, let's not introduce specific types for bucket IDs. It's not a domain logic, we don't have any specific operations defined on the bucket numbers so let's keep it simple.

src/main/java/io/tarantool/driver/api/space/options/crud/contracts/OperationWithFieldsOptions.java
```diff
@@ -11,17 +16,15 @@
...
     /**
      * Specifies list of fields names for getting only a subset of fields.
      * By default, all fields are returned.
      *
      * @param fields list of string field names
      * @return this options instance
      */
-    default T withFields(List<String> fields) {
-        addOption(FIELDS, fields);
+    default T withFields(FieldsValue fields) {
```
akudiyar, line 26: Here too, let's not introduce any new entities without strict necessity

src/main/java/io/tarantool/driver/api/space/options/crud/enums/CRUDOption.java
akudiyar: CRUD option is a good name however if take that we plan to split the CRUD-related code from the driver code. Let's discuss this if you wish

src/test/java/io/tarantool/driver/integration/proxy/options/ProxySpaceInsertOptionsIT.java
```diff
@@ -94,21 +95,22 @@ public void withTimeout() throws ExecutionException, InterruptedException {
         // with config timeout
         profileSpace.insert(tarantoolTuple).get();
         List<?> crudInsertOpts = client.eval("return crud_insert_opts").get();
-        assertNull(((HashMap) crudInsertOpts.get(0)).get("timeout"));
+        Assertions.assertNull(((HashMap<?, ?>) crudInsertOpts.get(0)).get(CRUDOption.TIMEOUT.getName()));
```
akudiyar, line 98: We may want to override `toString` perhaps to get rid of these `getName` calls

akudiyar (changes requested):
Overall: let's not introduce new entities (except enums) if the values do not have any special operations defined on them. IIf there is a strict necessity to do that, let's then
1) not make the user write `new` every time (use factory methods like `MyClass.of(value)`)
2) allow the user to use simple (built-in) types where it is possible

Anyway, any such changes are breaking so let's discuss them first. We will need to explain that in a doc and ideally add a memo/page in the wiki to show the users how to fix their code after the update.

Regarding the option names: having CRUD in them is logical, but let's therefore make them easily separable from the rest driver code to eventually make a new library package only for the tarantool/crud support.

nickkkccc:
I think I was able to get rid of the duplication of the CRUDOptions classes (identical to the ProxyOptions classes) by moving the options directly into the argument list.

src/main/java/io/tarantool/driver/api/space/options/BaseOptions.java
```diff
@@ -35,4 +39,13 @@ public void addOption(ProxyOption option, Object value) {
     public <T> Optional<T> getOption(ProxyOption option, Class<T> optionClass) {
         return Optional.ofNullable((T) resultMap.get(option));
     }
+
+    public Map<String, Object> asMap() {
```
ArtDu, line 43: Do we need cope of the map only for prechecking null values? If it's true we can add null checking in addOption method and remove copying

src/main/java/io/tarantool/driver/core/proxy/AbstractProxyOperation.java
```diff
@@ -2,7 +2,7 @@
 
 import io.tarantool.driver.api.SingleValueCallResult;
 import io.tarantool.driver.api.TarantoolCallOperations;
-import io.tarantool.driver.api.space.options.Options;
+import io.tarantool.driver.api.space.options.interfaces.Options;
```
ArtDu, line 5: If this change involves user API we need to add it as breaking change in changelog

ArtDu (commented):
We need to think how these changes modify cluster client working pipeline

CHANGELOG.md
```diff
@@ -4,9 +4,10 @@
 
 ### API changes
 
-- Add "mode" option for crud select operation ([#107](https://github.com/tarantool/cartridge-java/issues/107))
-- Change using of proxy client parameters (mode, rollback_on_error, stop_on_error) with enum classes ([#419](https://github.com/tarantool/cartridge-java/issues/419))
-
+- Add `"mode"` option for crud select operation ([#107](https://github.com/tarantool/cartridge-java/issues/107))
```
akudiyar, line 7: Just "select operation" perhaps? Is we are actually making unified API not only for tarantool/crud

src/main/java/io/tarantool/driver/api/TarantoolClient.java
```diff
@@ -4,7 +4,7 @@
 import io.tarantool.driver.api.connection.TarantoolConnectionListeners;
 import io.tarantool.driver.api.metadata.TarantoolMetadataOperations;
 import io.tarantool.driver.api.metadata.TarantoolMetadataProvider;
-import io.tarantool.driver.api.space.TarantoolSpaceOperations;
+import io.tarantool.driver.api.space.options.interfaces.TarantoolSpaceOperations;
```
akudiyar, line 7: Why do we need to make the packet names longer?

src/main/java/io/tarantool/driver/api/space/options/BaseOptions.java
akudiyar: I think we need an immutable map here

src/main/java/io/tarantool/driver/api/space/options/contracts/OperationWithBucketIdOptions.java
```diff
@@ -1,6 +1,8 @@
-package io.tarantool.driver.api.space.options;
+package io.tarantool.driver.api.space.options.contracts;
```
akudiyar, line 1: What does this new package name mean, what does it give to the users?

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
```diff
@@ -32,33 +32,22 @@ private DeleteProxyOperation(
...
-                .withTimeout(options.getTimeout())
-                .withBucketId(options.getBucketId())
-                .withFields(options.getFields())
-                .build();
-
-            List<?> arguments = Arrays.asList(spaceName, indexQuery.getKeyValues(), requestOptions.asMap());
 
             return new DeleteProxyOperation<>(
-                this.client, this.functionName, arguments, this.argumentsMapper, this.resultMapper);
+                this.client, this.functionName, new ArrayList<>(arguments.values()), this.argumentsMapper,
```
akudiyar, line 49: Why do we have that additional wrapping here? It will result in significant overhead for memory and GC time at least

src/main/java/io/tarantool/driver/core/proxy/contracts/package-info.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * Contains contract interfaces for proxy operation builders options.
+ *
+ * @author <a href="https://github.com/nickkkccc">Belonogov Nikolay</a>
+ */
+package io.tarantool.driver.core.proxy.contracts;
```
akudiyar, line 6: Do we really need that complex package structure? I haven't seen "contracts" packages in other libraries.

src/main/java/io/tarantool/driver/api/space/options/contracts/OperationWIthBatchSizeOptions.java
```diff
@@ -0,0 +1,34 @@
+package io.tarantool.driver.api.space.options.contracts;
+
+import io.tarantool.driver.api.space.options.enums.ProxyOption;
+import io.tarantool.driver.api.space.options.interfaces.Options;
+import io.tarantool.driver.api.space.options.interfaces.Self;
+
+import java.util.Optional;
+
+public interface OperationWIthBatchSizeOptions<T extends OperationWithTimeoutOptions<T>> extends Options, Self<T> {
```
akudiyar, line 9: Mistype in the class name

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
nickkkccc: Wrapping in enummap  was required to make unification across interfaces possible. If we don't use enummap, how can we connect the parts of the argument list in the correct number and order? After all, for each operation there are parts of the argument list that are different in composition and order.

CHANGELOG.md
nickkkccc: Good point.

src/main/java/io/tarantool/driver/core/proxy/enums/ProxyOperationArgument.java
```diff
@@ -0,0 +1,34 @@
+package io.tarantool.driver.core.proxy.enums;
+
+/**
+ * Enum class representing a parts of arguments from which the final list of arguments is combined. The class was
+ * introduced to combine interfaces with different numbers and types of compound arguments from which the final list
+ * of arguments is combined.
```
ArtDu, line 6: Also need to add warning message that order of parameters is very important. And that arguments combination creates a signature of function

src/main/java/io/tarantool/driver/api/TarantoolClient.java
nickkkccc: I wanted a more organized arrangement of interfaces and classes that implement those interfaces. A contract is a contract that interfaces-options expose to classes that implement it.

src/main/java/io/tarantool/driver/api/space/options/contracts/OperationWIthBatchSizeOptions.java
nickkkccc: Good point.

nickkkccc: Done.

CHANGELOG.md
nickkkccc: Done.

src/main/java/io/tarantool/driver/api/space/options/interfaces/Options.java
```diff
@@ -1,7 +1,8 @@
-package io.tarantool.driver.api.space.options;
+package io.tarantool.driver.api.space.options.interfaces;
 
-import io.tarantool.driver.api.space.options.enums.crud.ProxyOption;
+import io.tarantool.driver.api.space.options.enums.ProxyOption;
 
+import java.util.Map;
 import java.util.Optional;
 
 /**
```
ArtDu, line 8: I forgot about it in previous PR. 

Please add notion in javadoc

```
TODO: replace ProxyOption to general interface to have ability to use it not only in ProxyClient
```

src/main/java/io/tarantool/driver/api/space/options/interfaces/SelectOptions.java
```diff
@@ -0,0 +1,18 @@
+package io.tarantool.driver.api.space.options.interfaces;
+
+import io.tarantool.driver.api.space.options.contracts.OperationWIthBatchSizeOptions;
+import io.tarantool.driver.api.space.options.contracts.OperationWithBucketIdOptions;
+import io.tarantool.driver.api.space.options.contracts.OperationWithFieldsOptions;
+import io.tarantool.driver.api.space.options.contracts.OperationWithModeOptions;
+import io.tarantool.driver.api.space.options.contracts.OperationWithTimeoutOptions;
+
+/**
+ * Marker interface for space select operation options
```
ArtDu, line 10: Also add notion here

```
TODO: separate proxy options and cluster options
```

src/main/java/io/tarantool/driver/api/space/options/contracts/OperationWIthBatchSizeOptions.java
```diff
@@ -0,0 +1,34 @@
+package io.tarantool.driver.api.space.options.contracts;
+
+import io.tarantool.driver.api.space.options.enums.ProxyOption;
+import io.tarantool.driver.api.space.options.interfaces.Options;
+import io.tarantool.driver.api.space.options.interfaces.Self;
+
+import java.util.Optional;
+
+public interface OperationWIthBatchSizeOptions<T extends OperationWIthBatchSizeOptions<T>> extends Options, Self<T> {
```
ArtDu, line 9: Typo

```suggestion
public interface OperationWithBatchSizeOptions<T extends OperationWIthBatchSizeOptions<T>> extends Options, Self<T> {
```

ArtDu (changes requested):
Good job on improvement our complex API. I wrote some comments. It will be the best if you also create tickets about my notions in java doc

src/main/java/io/tarantool/driver/core/proxy/enums/ProxyOperationArgument.java
nickkkccc: Done.

src/main/java/io/tarantool/driver/api/space/options/interfaces/Options.java
nickkkccc: Done.

src/main/java/io/tarantool/driver/api/space/options/interfaces/SelectOptions.java
nickkkccc: Done.

ArtDu (dismissed):

nickkkccc:
@akudiyar, we decided not to make Map immutable for now. If necessary, we will change it in the next stages.

src/main/java/io/tarantool/driver/api/TarantoolClient.java
akudiyar: All this `.api.` package is about contracts. Adding more contract sub-layers here seems an overkill

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
akudiyar: I mean the `new ArrayList<>()` call first of all

akudiyar (changes requested):
Let's not introduce new packages like `interfaces` and `contracts` since we are already in the `api` package that means exactly that. Also think of the amount of changes the users will have to make in their code, will these changes be meaningful?

Also be careful about creating new objects on each call, `new ArrayList<>` for example is not necessary

src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java
nickkkccc: I realize that calling `new ArrayList<>()` creates intermediate collections. But it seems to me that the same collection is created before, which is later referenced by the "final" argument list:
https://github.com/tarantool/cartridge-java/blob/b743fe5de8a43b117e1ad20be273957a27d36bfb/src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java#L57-L64
Either I don't have enough experience to understand the difference between the two cases:
`Master case:`
https://github.com/tarantool/cartridge-java/blob/b743fe5de8a43b117e1ad20be273957a27d36bfb/src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java#L57-L64
`Commit case:`
https://github.com/tarantool/cartridge-java/blob/670297e4143f55f04a3d9217a269c5489e66b106/src/main/java/io/tarantool/driver/core/proxy/DeleteProxyOperation.java#L47-L53

akudiyar (changes requested):
See my comment above - let's discuss that verbally if you wish

PR #439, opened 2023-11-12
Replace streams with loops in DefaultMessagePackMapper

akudiyar:
`toValue` and `fromValue` methods in DefaultMessagePackMapper are hot spots as they are called on each request. This code was first written using Java streams, but while it allowed to make the code more readable, the streams are not very well-suitable for this task. The Java streams work better when we can execute the chained lambda functions in parallel on an executor, but are worse when replacing the normal `for` loops - they add significant overhead because of a lot of created auxiliary classes and objects in runtime.

Now as the DefaultMessagePackMapper code is stable we can replace the stream-based code with more optimized loop-based code.

### DIFF
--- a/src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
+++ b/src/main/java/io/tarantool/driver/api/MessagePackMapperBuilder.java
@@ -7,6 +7,7 @@ import io.tarantool.driver.mappers.converters.ValueConverter;
 import org.msgpack.value.Value;
 import org.msgpack.value.ValueType;
 
+import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


akudiyar:
Waits for #438 to be merged.

ArtDu (commented):
LGTM, only one request: add lines in changelog for this PR and for the prev one https://github.com/tarantool/cartridge-java/pull/438

akudiyar:
Changelog line added in https://github.com/tarantool/cartridge-java/pull/453

I guess we can merge this PR as it is.

ArtDu (approved):

PR #483, opened 2025-09-16
Use JDK 11 with  release option for JDK 8

akudiyar:
Without this option, the compiled binaries are different from those generated by the JDK 8 and this breaks ancient libraries like Spark 2.4.x when it is used together with the newer driver versions.


akudiyar:
TODO:

- [ ] Migrate test container Dockerfile to Ubuntu

